PR #62, opened 2021-03-30
Eagle 300

M-Wicenec:


src/Eagle.ts
```diff
@@ -44,6 +44,7 @@ import {Edge} from './Edge';
 import {Field} from './Field';
 import {FileInfo} from './FileInfo';
 import {Setting} from './Setting';
+import { collapseTextChangeRangesAcrossMultipleVersions } from "typescript";
```
james-strauss-uwa, line 47: Do we need this import? The function "collapseTextChangeRangesAcrossMultipleVersions" doesn't seem to be used in Eagle.ts. If not required, you could remove it. If it is required, then please add a comment with the reason and/or a URL where the fix was found.

src/Eagle.ts
```diff
@@ -2585,7 +2586,37 @@ export class Eagle {
         return true;
     }
 
-   
+    spinCollapseIcon = (item:any, e:JQueryEventObject) => {
```
james-strauss-uwa, line 2589: There are just a few "style" issues here. The code is fine, it just doesn't exactly fit with the existing style

1. The variable name "CollapseTarget" begins with a capital letter, I tend to use a lower-case first letter in the camel case for this Typescript
2. Most lines do not end with a semi-colon, which seems to be valid Typescript, but I think we should keep it consistent.
3. The indenting of line 2604 has one tab too many.
4. Two line gap on lines 2597-8 should just be one line
5. The name of the new class you have defined "ClosedIcon" should start with a lower-case letter
6. Lines 2604-2608 and 2612-2616 seem to be repeated. I think it would be better to remove this code from both the if and else blocks, and just have one copy outside (after) the if-else

src/Palette.ts
```diff
@@ -126,7 +126,9 @@ export class Palette {
     // TODO: this should return different icons based on whether the palette is currently expanded or collapsed
     //       but at the moment, that expand/collapse state is stored internally within bootstrap and is not available here
     getCollapseIcon = () : string => {
-        return "keyboard_arrow_down";
+        // if($("#collapse0").hasClass("show")){
```
james-strauss-uwa, line 129: It looks like you could remove the commented lines here.

static/base.css
```diff
@@ -298,4 +298,14 @@ nav.navbar .introjs-helperLayer {
 
 .nodeTitle{
     pointer-events:none;
+}
+
+.ClosedIcon{
```
james-strauss-uwa, line 303: Use camel case beginning with a lower-case letter for the class name here

templates/palettes.html
```diff
@@ -1,7 +1,7 @@
 <div id="paletteList">
     <div data-bind="foreach: palettes">
         <div class="card">
-            <div class="card-header" data-toggle="collapse" aria-expanded="true" data-bind="attr:{id:'palette' + $index(), 'data-target':'#collapse' + $index(), 'aria-controls':'collapse' + $index()}">
+            <div class="card-header" data-toggle="collapse" aria-expanded="true" data-bind="attr:{id:'palette' + $index(), 'data-target':'#collapse' + $index(), 'aria-controls':'collapse' + $index()}, event: {click: $root.spinCollapseIcon}">
```
james-strauss-uwa, line 4: You've used the "event" binding here, which is fine. There is also a binding in knockoutJS called "click", specifically for the click event. So you could write this as data-bind="click: $root.spinCollapseIcon", instead of data-bind="event: {click: $root.spinCollapseIcon}".

But this comment is just to remind you , no need to change the code.

james-strauss-uwa (changes requested):
Looks great and works well, thanks.

Again I have just left a few comments regarding minor clean up and style issues.

Thanks.

src/Eagle.ts
M-Wicenec: i dont remember adding this line. ill ask Andreas if its from him

templates/palettes.html
M-Wicenec: I remember having it as "click: $root.spinCollapseIcon" initially but it triggered the function on creation of the element, meaning each time you selected or created a node it would trigger the function and throw an error.

james-strauss-uwa: I have troubles with this issue too, quite frequently.

In my experience, if you use "click: $root.spinCollapseIcon()" (note the trailing brackets), then the function will be executed at the time of binding.
If you use "click: $root.spinCollapseIcon" (no trailing brackets), then the function will be executed when the click event occurs.

Having said that, you'll see instances throughout the code where I have wrapped the function call in an anonymous function. Something like: "click: function(){$root.spinCollapseIcon();}" just to ensure the function is executed at event time (or if the function needs some arguments).

PR #72, opened 2021-04-08
Eagle 305

M-Wicenec:


src/Eagle.ts
```diff
@@ -84,6 +84,7 @@ export class Eagle {
 
     static nodeDropped : Element;
     static nodeDropLocation = {x:0, y:0}; // if this remains x=0,y=0, the button has been pressed and the getNodePosition function will be used to determine a location on the canvas. if not x:0, y:0, it has been over written by the nodeDrop function as the node has been dragged into the canvas. The node will then be placed into the canvas using these co-ordinates.
+    static nodeInspectorCount : number = 0; // this is 0 if all menus collapsed and counts how many are open
```
james-strauss-uwa, line 87: I think it is important to know when the inspectorCount is zero, and when it is non-zero. Those are the two cases in which the code needs to behave differently. So I ended up changing this to a boolean.

src/Eagle.ts
```diff
@@ -408,13 +409,18 @@ export class Eagle {
     }
 
     setSelection = (rightWindowMode : Eagle.RightWindowMode, selection : Node | Edge) : void => {
-        //console.log("eagle.setSelection()", Utils.translateRightWindowModeToString(rightWindowMode), selection);
+        //set amount of collapsed node menus to default (0)
 
+        $(".nodeInspectorCollapseAll").collapse("hide");
+        $(".nodeMenuIndicator").addClass("closedIcon");
```
james-strauss-uwa, line 415: These calls to addClass and removeClass were what initially made me think of trying a knockout-like approach. Knockout has a data-bind specifically for this purpose.

data-bind="css: {className: conditional-to-evaluate}".

If the conditional evaluates as true, then knockout adds the class (className) to the element, otherwise it removes the class.

src/Eagle.ts
```diff
@@ -2640,11 +2646,25 @@ export class Eagle {
         if (triggerClass){
```
james-strauss-uwa, line 2646: This function performs the same behaviour (spinCollapseIcon) for a few different element types.

I found it a bit difficult to read and understand all the different cases.

I think triggerClass might be better named something like "isTranslationToggle".

src/Eagle.ts
```diff
@@ -2640,11 +2646,25 @@ export class Eagle {
...
             //this is for setting toggle icons in the translation menu, as the collapse functions differently and the content is nested differently.
             //the class "closedIcon" turns the collapse arrow icon by 270 degrees and is being toggled depending on the current state of the collapse.
-            $(".translationToggle").addClass("closedIcon")
+            $(".translationToggle").addClass("closedIcon");
             var toggleState = collapseTarget.parent().parent().parent().children(".collapse").hasClass('show');
```
james-strauss-uwa, line 2650: I think collapseTarget is the "i" tag (the icon), so that might be a better name for the collapseTarget variable.

Here we use 3 parent() calls, lower down we use 2 parent() calls. It would be nice if we could standardise the arrangement of the HTML, so that the collapseTarget and the collapsable object always have the same relative arrangement.

src/Eagle.ts
```diff
@@ -2640,11 +2646,25 @@ export class Eagle {
...
+            //this is for the global node inspector collapsable counter to keep the collapse all button state synced
+            if(collapseTarget.parent().parent().children(".collapse").hasClass("nodeInspectorCollapseAll")){
+                if(toggleState){
+                    Eagle.nodeInspectorCount --;
+                }else{
+                    Eagle.nodeInspectorCount ++;
+                }
+
+                if (Eagle.nodeInspectorCount === 0){
+                    $("#nodeInspectorHeading").find('i').first().addClass("closedIcon");
```
james-strauss-uwa, line 2663: Again we are using addClass/removeClass here, which made me think that making Eagle.nodeInspectorCount a knockout observable and then data-binding the "closedIcon" class of the "i" tag to it might be a better approach.

src/Eagle.ts
```diff
@@ -2654,6 +2674,28 @@ export class Eagle {
         }
     }
 
+    toggleAllNodeMenus = () => {
+        //toggle all node inspector collapsables functionality
+        if(Eagle.nodeInspectorCount != 0){
+            $(".nodeInspectorCollapseAll").collapse("hide");
+            $(".nodeMenuIndicator").addClass("closedIcon");
+            $("#nodeInspectorHeading").find('i').first().addClass("closedIcon");
```
james-strauss-uwa, line 2682: We see this addClass/removeClass anti-pattern again here

templates/node_inspector.html
```diff
@@ -1,12 +1,27 @@
 <div id="nodeInspector" class="container px-0">
     <!-- ko if: selectedNode() != null -->
 
-        {% include 'node_inspector_attributes.html' %}
-
+        <h5 class="card-header" id="nodeInspectorHeading" data-bind="click: $root.spinCollapseIcon, click: $root.toggleAllNodeMenus">
```
james-strauss-uwa, line 4: I think this section of HTML could be split out into a new file "node_inspector_header.html" or similar. Just so that we follow the pattern of each section of the node_inspector having its own file.

Lots of small files is usually preferable to one large file for this sort of thing.

templates/node_inspector.html
```diff
@@ -1,12 +1,27 @@
...
+            <i class="material-icons interactive closedIcon" data-toggle="tooltip" data-placement="top" data-html="true" data-bind="attr:{title:'Expand/Collapse All'}">keyboard_arrow_down</i>
+            <!-- ko if: selectedNode().isLocked() -->
+            <i class="material-icons interactive" data-toggle="tooltip" data-placement="top" data-html="true" data-bind="attr:{title:'readonly'}">lock</i>
+            <!-- /ko -->
+            <!-- ko ifnot: selectedNode().isLocked() -->
+            <i class="material-icons interactive" data-toggle="tooltip" data-placement="top" data-html="true" data-bind="attr:{title:'readwrite'}">lock_open</i>
+            <!-- /ko -->
+            <span data-bind="text: selectedNode().getName()"></span>
+        </h5>  
+            <h4 class="nodeInspectorSubTitle" id="firstNodeInspectorSubtitle">Key: <span data-bind="text: selectedNode().getKey()"></span></h4>
```
james-strauss-uwa, line 14: The indentation seems inconsistent here.

templates/node_inspector.html
```diff
@@ -45,20 +60,22 @@
         {% include 'node_inspector_graph_comment.html' %}
         <!-- /ko -->
 
-        <div class="row mt-1">
+        {% include 'node_inspector_hidden_bs_trigger.html' %}
+    </div>
```
james-strauss-uwa, line 64: Indentation

templates/node_inspector_attributes.html
```diff
@@ -1,112 +1,25 @@
 <div class="card">
```
james-strauss-uwa, line 1: I think this file is not included (in node_inspector.html) anymore. So I think we can remove the whole file.

templates/node_inspector.html
```diff
@@ -45,20 +60,22 @@
         {% include 'node_inspector_graph_comment.html' %}
         <!-- /ko -->
 
-        <div class="row mt-1">
+        {% include 'node_inspector_hidden_bs_trigger.html' %}
```
james-strauss-uwa, line 63: In my branch (eagle-305-2), I was able to remove the need for this hidden trigger.

It's a little complicated to explain how here in text, but hopefully it is obvious if you take a look at the code. Otherwise we can discuss it on Thursday.

james-strauss-uwa (commented):
I think these changes work really well, and this pull request has been valuable in working out what needs to be done. On reflection though, some of the changes you made, which we both acknowledged at the time were a little hack-ish, made me wonder if doing things in a more knockout-like style would result in a cleaner solution.

To explore those ideas, I branched from eagle-305 as eagle-305-2, which is a more knockout-like style. The HTML and CSS hardly differs from eagle-305, but the javascript has a few changes. I think I like some of the aspects of the approach in eagle-305-2, but there are also some weaknesses.

I had hoped to address some of the weaknesses before showing you, but it has already been a few days, so I think I should just get your feedback sooner. Perhaps we can come up with some combination of the two approaches.

PR #75, opened 2021-04-12
Eagle 305 2

james-strauss-uwa:
A knockout-style approach to EAGLE-305.

We add one observable boolean to Eagle (nodeInspectorCollapsed) which stores the state of the outer collapse icon
And 3 (at this stage, more in future) observable booleans to each Node, which store the collapsed/expanded state of each section of the node inspector.

This absolutely has weaknesses, so looking for your feedback on this kind of approach.


M-Wicenec:
The one issue that stood out is that when a collapse is clicked in quick succession (while bootstrap is still collapsing the section) this knockout code gets confused between states. Bootstrap has a class on the collapsable section during transition called "collapsing"
The fix we discussed earlier was to use that transition class to exit the function and effectively ignore the second click.
Otherwise i think this is a cleaner solution since it doesnt require a hidden object

PR #125, opened 2021-07-01
Eagle 385

awicenec:
Although we have abandoned the related ticket, there had been a few changes of the documentation and also the default sphinx theme in here, which actually led to moving the docs to readthedocs.


james-strauss-uwa (approved):
It's too large to review in detail. But the results on eagle-dlg.readthedocs.io look good.

PR #149, opened 2021-07-29
Add 'toRef' and 'fromRef' (where appropriate) to edges in AppRef format

james-strauss-uwa:


awicenec:
Definitely a step forward. Graph is already translating, albeit wrongly.

PR #150, opened 2021-07-30
Fix Palette Uploading

calgray:
I noticed a small bug with a simple correction when uploading palette files where the `uploadedPaletteFile` element id does not exist. Since uploading graphs works fine this change simply updates the palette upload to follow the same pattern. Also checked no console errors occur after uploading.


james-strauss-uwa (approved):
Great, thanks

PR #155, opened 2021-08-10
LIU-117 Add PlasmaFlight Drop

calgray:
This change adds a new PlasmaFlight data drop that has a fallback to looking up plasma objects over a flight server when they don't exist locally. Further description is in https://github.com/ICRAR/daliuge/pull/58

image


james-strauss-uwa (approved):
Looks great, thanks. Great icon too.

PR #180, opened 2021-09-02
LIU-116 Support EndDrops

calgray:
With the recent addition of service nodes and node skipping behavior there are some cases now for use of EndDrops with changes from https://github.com/ICRAR/daliuge/pull/63

image

### DIFF
--- a/src/LogicalGraph.ts
+++ b/src/LogicalGraph.ts
@@ -515,7 +515,7 @@ export class LogicalGraph {
         newNode.setEmbedKey(null);
 
         // convert start of end nodes to data components
-        if (newNode.getCategory() === Eagle.Category.Start || newNode.getCategory() === Eagle.Category.End) {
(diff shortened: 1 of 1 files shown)


src/LogicalGraph.ts
```diff
@@ -515,7 +515,7 @@ export class LogicalGraph {
         newNode.setEmbedKey(null);
 
         // convert start of end nodes to data components
-        if (newNode.getCategory() === Eagle.Category.Start || newNode.getCategory() === Eagle.Category.End) {
+        if (newNode.getCategory() === Eagle.Category.Start) {
```
james-strauss-uwa, line 518: It looks like the "Eagle.Category.End" case of the switch within this if-block becomes redundant here.

We can probably remove the switch too, and just always use code in the "Eagle.Category.Start" case.

james-strauss-uwa (approved):
Looks good, happy for you to merge this, thanks.

PR #184, opened 2021-09-03
Eagle 428 2

M-Wicenec:


M-Wicenec:
Up to you if you would like to do some performance tests, to check if this change significantly impacts graph rendering performance. these arrowheads have to be redrawn when the svg is updated. it happens quite a lot when dragging a node

src/bindingHandlers/graphRenderer.ts
```diff
@@ -118,44 +119,29 @@ function render(graph: LogicalGraph, elementId : string, eagle : Eagle){
...
-        .attr("orient", "auto");
-
-    grey_arrowhead
-        .append("path")
-        .attr("d", "M 0 0 L 10 5 L 0 10 z")
-        .attr("stroke", "none")
-        .attr("fill", LINK_COLORS['LINK_DEFAULT_COLOR']);
+    //generating defs from colors array
+    Object.keys(LINK_COLORS).forEach(function (value, i) {
+        const x = defs
```
james-strauss-uwa, line 128: Maybe a better variable name here, instead of 'x', use 'arrowhead' or 'marker' or 'def'

src/bindingHandlers/graphRenderer.ts
```diff
@@ -2919,10 +2905,14 @@ function render(graph: LogicalGraph, elementId : string, eagle : Eagle){
             normalColor = LINK_COLORS['LINK_WARNING_COLOR'];
             selectedColor = LINK_COLORS['LINK_WARNING_SELECTED_COLOR'];
         }
-
         return eagle.objectIsSelected(edge) ? selectedColor : normalColor;
     }
 
+    function getArrowType(edge: Edge, index: number) {
```
james-strauss-uwa, line 2911: Perhaps modify this name from 'getArrowType' to something that follows the existing pattern of function names. 'edgeGetEndMarkerUrl' or 'edgeGetArrowheadUrl'.

src/bindingHandlers/graphRenderer.ts
```diff
@@ -2919,10 +2905,14 @@ function render(graph: LogicalGraph, elementId : string, eagle : Eagle){
...
+        const x = edgeGetStrokeColor(edge, index)
```
james-strauss-uwa, line 2912: Used x here too.

james-strauss-uwa (approved):
Just make those small changes and then you can merge

PR #203, opened 2021-10-04
Eagle 327 2 (bootstrap update and component parameter input changes)

M-Wicenec:
It might be a good idea for one of you to play around with the interface and look for bugs i may have missed. 
If you do, keep an eye out for;
- non working modals
- any weird styling issues
- tooltips not loading properly
- collapsable menus not working 
- any navbar weirdness


james-strauss-uwa:
I'm happy for you to remove the old bootstrap code instead of commenting it out. For example main.ts line 93

There are a few console.log() debug messages that we can clean up or remove

When I click on a boolean param, the value changes from true/false, but the background is still filled with the blue color. I think I saw on your computer that the background color should be blue when true, and white when false.

When I select a node, then click the "add selected node to palette" button in the heading bar of the node inspector, the modal that comes up is the "Destination Palette" modal. It has a drop down selector and a text input. There is no spacing between the two elements, which is a minor issue. I think there should be a small space.

When I did the above, then created a new palette as the destination, the new palette appears in the palette list on the left, but the "save to disk" and "save to git" buttons in the header are not styled correctly, they have white backgrounds and outlines.

The "Add Custom Repository" modal has inconsistent spacing between the three items. To view this click the "add custom repository" button at the bottom of the repository list.


In the navbar "Graph  > Git Storage > Save Graph As" pops up a modal , and the spacing in that modal is inconsistent

PR #221, opened 2021-10-19
small bugfix repository eject button was always showing

M-Wicenec:
not sure about the package-lock.json changes. where do they come from and should i be concerned when the file has changes?


james-strauss-uwa (approved):
package.json is where the top-level dependencies are recorder

package-lock.json also records all the dependencies of those dependencies

I think it is only modified when you use "npm install". It is recommended that the files be added to git.

PR #261, opened 2021-11-29
Eagle 556

james-strauss-uwa:
A new approach for handling tooltips throughout the code. It defines a knockout "custom binding", so that you can do:

`<span data-bind="eagleTooltip:'This is the tooltip text'">blah</span>`

By default, the custom binding will place the data-bs-toggle='tooltip', data-bs-placement='right' and data-html='true' attributes on the bound element. However, if data-bs-placement is already specified in the template HTML, that attribute will not be overwritten.

Thankfully, this approach removes the need to call reloadTooltips, and solves the timing issues that forced us to call setTimeout() to leave enough time for the HTML to be rendered. Now the init() and update() functions in the custom binding happen at the correct time by default.


M-Wicenec (approved):
all the tooltips seem to be updating and loading properly!

PR #289, opened 2022-01-10
Adds support for ParameterSet Drops

pritchardn:
A straightforward change adding support for a new ParameterSet drop.


awicenec:
This requires some additional changes in order to get the Category and Type corrected to also fix the color of that node. In particular ParameterSet has to be defined in Eagle.ts in the cData structure (lines 4295 ff). We also need to define an icon for this new type. We'll create a new ticket to tackle these issues and assign it to Moritz.

PR #290, opened 2022-01-10
LIU-216 Add shared memory size param

calgray:
After some testing with plasma stores I noticed that writing to plasma starts to fail when objects exceed the default docker shared memory size (even though the plasma store accepts a memory argument bigger than this). In Daliuge 2.1.0 I've added this extra argument to docker containers and tested that large objects can now be written to plasma.


awicenec:
Simple addition. All good.

PR #293, opened 2022-01-11
Yan 854 -- SharedMemory DROPs

pritchardn:
This merge adds another component - A shared Memory DROP (relates to [this sibling DALiuGE merge](https://github.com/ICRAR/daliuge/pull/92).

This drop is functionally equivalent to a standard Memory drop and is therefore treated as such, but differs in its underlying implementation.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -4206,7 +4206,7 @@ export class Eagle {
             const nodePosition = newNode.getPosition();
 
             // build a list of ineligible types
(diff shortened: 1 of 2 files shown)


awicenec:
While the implementation is fine, I think we should now consider whether to remove 3.7 support all together (probably easiest) or have some additional high-level check in daliuge to catch this incompatibility for shared memory drops if they had been requested in a graph.

PR #295, opened 2022-01-19
Liu 220 - EnvironmentVariable DROP

pritchardn:
Adds a new generic component for storing globally available parameters.
This should be a minimal edition, some tests however may need re-working from the looks of things however.


james-strauss-uwa:
If it is a hassle to get testcafe running locally, I could look into the reason for the test failures.

pritchardn:
Oops, many thanks for picking up on that one

static/template.palette
```diff
@@ -825,6 +825,32 @@
             "inputPorts": [],
             "readonly": true
         },
+        {
+            "category": "EnvironmentVars",
```
james-strauss-uwa, line 829: Sorry, I think I changed this to "EnvironmentVariables" in Eagle.ts

That might have been the cause of the conflicts with the master branch.

james-strauss-uwa (approved):
I'll just change "EnvironmentVars" to "EnvironmentVariables", then it's ready to merge.

PR #336, opened 2022-03-08
settings are now cancelable

M-Wicenec:


src/Eagle.ts
```diff
@@ -119,34 +119,34 @@ export class Eagle {
...
-                new Setting("Enable Performance Display", "Display the frame time of the graph renderer", Setting.Type.Boolean, Utils.ENABLE_PERFORMANCE_DISPLAY, false),
-                new Setting("Use Simplified Translator Options", "Hide the complex and rarely used translator options", Setting.Type.Boolean, Utils.USE_SIMPLIFIED_TRANSLATOR_OPTIONS, true),
-                new Setting("Show File Loading Warnings", "Display list of issues with files encountered during loading.", Setting.Type.Boolean, Utils.SHOW_FILE_LOADING_ERRORS, false),
-                new Setting("Graph Zoom Divisor", "The number by which zoom inputs are divided before being applied. Larger divisors reduce the amount of zoom.", Setting.Type.Number, Utils.GRAPH_ZOOM_DIVISOR, 1000)
+                new Setting("Confirm Discard Changes", "Prompt user to confirm that unsaved changes to the current file should be discarded when opening a new file, or when navigating away from EAGLE.", Setting.Type.Boolean, Utils.CONFIRM_DISCARD_CHANGES, true,''),
```
james-strauss-uwa, line 122: It seems like we are calling all these Settings constructors with '' as the oldValue. I think this makes sense, we will always use '', we can't know any better value to set. So maybe we can just leave out the oldValue param in the constructor, and assume it will always be the empty string.

src/Modals.ts
```diff
@@ -200,9 +200,38 @@ export class Modals {
 
         // #settingsModal - showSettingsModal()
         $('#settingsModal').on('shown.bs.modal', function(){
+            $('#settingsModal').data('completed', false);
+            eagle.copyCurrentSettings()
             $('#settingsModalAffirmativeButton').focus();
         });
 
+        $("#settingsModalNegativeButton").on('click', function(){
+            eagle.cancelSettingChanges()
```
james-strauss-uwa, line 209: I think this case will be automatically handled by the code on line 217.

If the user clicks the negative button, the 'hidden.bs.modal' event will occur, and 'completed' will be false. So eagle.cancelSettingChanges() will get called anyway.

src/Setting.ts
```diff
@@ -10,14 +10,16 @@ export class Setting {
     private type : Setting.Type;
     private key : string;
     private defaultValue : any;
+    oldValue : any;
```
james-strauss-uwa, line 13: This variable can probably be private.

src/Setting.ts
```diff
@@ -10,14 +10,16 @@ export class Setting {
...
 
-    constructor(name : string, description : string, type : Setting.Type, key : string, defaultValue : any){
+    constructor(name : string, description : string, type : Setting.Type, key : string, defaultValue : any, oldValue:any){
```
james-strauss-uwa, line 15: You could remove the "oldValue" parameter here, and just assume it is "".

src/Setting.ts
```diff
@@ -10,14 +10,16 @@ export class Setting {
...
         this.name = name;
         this.description = description;
         this.type = type;
         this.key = key;
         this.value = ko.observable(defaultValue);
         this.defaultValue = defaultValue;
+        this.oldValue = oldValue;
```
james-strauss-uwa, line 22: And here just use:

this.oldValue = "";

james-strauss-uwa (changes requested):

james-strauss-uwa (approved):

PR #351, opened 2022-04-22
Eagle 681 2

M-Wicenec:


src/Eagle.ts
```diff
@@ -68,6 +68,7 @@ export class Eagle {
 
     static parameterTableType : ko.Observable<string>;
     static parameterTableSelectionParent : ko.Observable<Field>; // row in the parameter table that is currently selected
+    static parameterTableSelectionParentId : ko.Observable<number> // id of the selected field
```
james-strauss-uwa, line 71: Maybe "parameterTableSelectionParentIndex" is a better name for this variable.

src/Eagle.ts
```diff
@@ -2445,13 +2447,23 @@ export class Eagle {
             Utils.showNotification("Error", "No Node Is Selected", "warning");
         }else{
             if (tableType === 'component'){
```
james-strauss-uwa, line 2449: This is more of a style-related thing than a correctness thing, but I think this sort of comparison is error-prone. It is too easy to misspell 'component'.

We have an enumerated type (Eagle.FieldType) which has two possible values (Field, and ApplicationParam). I'd be tempted to use this instead of the literal string 'component'.

You could also make the tableType parameter a Eagle.FieldType instead of a string.

src/Eagle.ts
```diff
@@ -3241,12 +3256,34 @@ export class Eagle {
...
 
-    addEmptyField = (node:Node) : void => {
-        node.addEmptyField(this.selectedNode())
-    }
+    addEmptyTableRow = () : void => {
+        var fieldIndex
+        if(Eagle.parameterTableSelectionParentId() != -1){
+        // A cell in the table is selected well insert new row instead of adding at the end
+            fieldIndex = Eagle.parameterTableSelectionParentId()+1
+            if(Eagle.parameterTableType() === 'component'){
```
james-strauss-uwa, line 3264: Maybe Eagle.parameterTableType could be of type Eagle.FieldType, instead of string.

If you want to reset Eagle.parameterTableType to neither 'component' or 'application', you would normally set it to ''. But if you make the change to use a Eagle.FieldType instead, you may want to add a Eagle.FieldType.Unknown value to the enum.

src/Eagle.ts
```diff
@@ -3241,12 +3256,34 @@ export class Eagle {
...
+            }else{
+                //argument table
+                this.selectedNode().addEmptyArg(fieldIndex)
+            }
+        }else{
+        //no cell selected, add new row at the end
+            if(Eagle.parameterTableType() === 'component'){
+                //component table
+                this.selectedNode().addEmptyField(-1)
+                fieldIndex = this.selectedNode().getFields().length -1
```
james-strauss-uwa, line 3276: Could using eagle.currentParamsArray() simplify things here?

src/Node.ts
```diff
@@ -1145,12 +1150,20 @@ export class Node {
         this.fields()[0].setValue(e.value);
     }
 
-    addEmptyField = (selectedNode:Node) :void => {
-        selectedNode.addField(new Field("New Parameter", "", "", "", "", false, Eagle.DataType.String, false, [], false));
+    addEmptyField = (index:number) :void => {
+        if(index === -1){
```
james-strauss-uwa, line 1154: The "new Field(...)" code is the same in both cases, so perhaps create the new field once outside the if, and then using it inside the if becomes cleaner.

src/Eagle.ts
```diff
@@ -2445,13 +2447,23 @@ export class Eagle {
             Utils.showNotification("Error", "No Node Is Selected", "warning");
         }else{
             if (tableType === 'component'){
+                if (!this.selectedNode().canHaveParameters()){
+                    Utils.showNotification("Error", "This Node Cannot Have Parameters", "warning");
+                    return
+                }
                 $("#parameterTableModalTitle").html("Component Parameter Table")
+                $("#parameterTableModalAddParameterButton").html("Add Parameter")
```
james-strauss-uwa, line 2455: This looks a little unusual to me. We'd normally try to do things the "knockout way" and update the html using a data-bind in the html code, rather than update it using jquery.

It would be something like:

<div data-bind="html: Eagle.parameterTableType() === Eagle.FieldType.Field ? 'Add Argument' : 'Add Parameter'">

It's about the same amount of code, so not a huge difference either way. But I guess it's more knockout-y to use the data-bind in the html code.

james-strauss-uwa (approved):
Just a few small recommendations. Otherwise looks great.

PR #354, opened 2022-04-28
Eagle 523

james-strauss-uwa:
Added logical graph undo and redo. Default buffer size is 10 steps. Moved zoom in/out to use +/- to allow undo/redo to use z/shift-z.


M-Wicenec (changes requested):
just the keyboard shortcut modal issues i commented on. the rest looks great

PR #363, opened 2022-05-05
Eagle 719 - remove duplicate DOM elements with the same id (tableInspectorSelection, tableInspectorValue)

james-strauss-uwa:
The motivation here was to get rid of two warnings in Chrome about two DOM elements sharing each id (tableInspectorSelection, tableInspectorValue). The elements were defined in templates/modals.html. One meant to be used if Eagle.parameterTableSelection was null, and one if not null. In each case, I combined these two elements into one, and moved the null-check into the javascript.

I also made a small change to the type definition of Eagle.parameterTableSelection, it seems to store a string instead of a Field.

Finally, I changed the way Field.select() is called, just passing it a string instead of the whole knockout observable object.


M-Wicenec (approved):
looks, good. im no longer getting any console warnings due to duplicate ids. the palette ones seem to have resolved themselves as well.

PR #366, opened 2022-05-12
Eagle 721 - Combine expert mode and advanced editing settings

james-strauss-uwa:
Added two new attributes to the Settings class (displayFunc and toggleFunc).

- displayFunc is a function that is run to determine whether the setting should appear in the list of settings.
- toggleFunc is a function that is run whenever the setting is changed by the user.

This allows us to hide several "advanced editing" and "workaround" settings, until "expert mode" is enabled.

Then, when expert mode is enabled, we run a toggleFunc which enabled ALL the "advanced editing" settings automatically.

Also, I added one new setting ("Show DALiuGE runtime parameters") that hides/shows the list of component parameters common to DALiuGE. Previously, this behaviour was controlled by the "expert mode" setting, but now it has its own.

All "expert mode" does is hide/show all the "advanced editing" settings, enabling them all by default when it is enabled, and disabling them all when it is disabled.

### DIFF
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -40,7 +40,7 @@ export class Config {
         "simple"
     ];
 
-    static readonly EXPERT_MODE_PARAMETER_NAMES = [
+    static readonly DALIUGE_PARAMETER_NAMES = [
         "data_volume",
         "execution_time",
         "num_cpus",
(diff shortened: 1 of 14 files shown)


james-strauss-uwa:
OK, it should be ready to go now. Please check once more, thanks.

james-strauss-uwa:
Could you check my fix for the "reset defaults" issue in the settings modal? Thanks.

M-Wicenec:
looks all good now

PR #405, opened 2022-06-24
Eagle 775

james-strauss-uwa:
This work was motivated by a problem with EAGLE that occurred when a user wanted to add a port to a data component. Since data components are not allowed to be given "application arguments", then there was no way to promote one of those into a port.

Instead, we decided that users would add a component parameter, as was already allowed, but they we'd modify EAGLE to allow those component parameters to be promoted to ports.

This also meant that we needed the parameters table to show everything (component parameters/application arguments/input ports/output ports) even when launched from the button on the header of the component parameters section.

So this removed the need for two different "modes" of the parameter table. So I've removed the button from the "application arguments" section, since sometimes that is invisible. And the single button displays the table in a single mode, where everything is shown.

Internally, all four things (component parameters/application arguments/input ports/output ports) are now stored in a single array Node.fields. But input and output is the same as before, the JSON has 4 different arrays.

This change touched a lot of areas, so please take a look at the behaviour of the parameter table, and at promoting fields into ports and back again.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -2453,18 +2453,17 @@ export class Eagle {
     }
 
     getCurrentParamReadonly = (index: number) : boolean => {
-
         if(Eagle.selectedLocation() === Eagle.FileType.Palette){
             if(Eagle.allowPaletteEditing()){
                 return false;
             }else{
-                return this.selectedNode().getFieldReadonly(index);
+                return this.selectedNode().getFields()[index].isReadonly();
             }
         }else{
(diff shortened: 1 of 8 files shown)


james-strauss-uwa:
I spoke to @M-Wicenec about this, he's happy with the behaviour, but pointed out two possible improvements. First to move the "open table" button to a more suitable place. Second, to add filters to the table, to display subsets of the parameters, for example, just ports.

We've created new issues for these suggestions. I'll treat them as additional features, and merge this work first.

PR #406, opened 2022-06-28
Eagle 761

james-strauss-uwa:
This switches to using a string for the "type" of each field (component parameter/application argument/input port/output port) instead of an enumerated type.

We need a much richer type system in order to describe all the various output types from components.

There are still several "known" types (Integer, Float, Boolean, String etc). But the user can specify any string.

Perhaps check the functionality surrounding creating new fields, changing types, both in the edit field modal, and in the parameters table.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -3436,6 +3436,17 @@ export class Eagle {
         }
     }
 
+    editFieldDropdownClick = (newType: string, oldType: string) : void => {
+        console.log("editFieldDropdownClick", newType, oldType);
+
+        // check if the types already match, therefore nothing to do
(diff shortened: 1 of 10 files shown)


james-strauss-uwa:
This is ready to review @M-Wicenec

M-Wicenec (changes requested):
The new functionality works well. the only thing i noticed broken is in the edit field modal the value input type doesnt immediately change to an appropriate one. 
eg. when switching from string to boolean the value input field remains a textbox instead of switching to a toggle button. select and boolean worked previously, however integer and float dont seem to work on previous versions either

PR #408, opened 2022-06-30
Eagle 723

james-strauss-uwa:
In this change, when the user drags an edge between two application nodes, the intermediate data node is always automatically created as a Memory node. It is then up to the user to click the "edit node category" button and choose a new category, if they want to use something else, like a File or Plasma component.

Previously, users would always be prompted immediately to decide what category of data component was required.

The list of possible categories to which an existing node can be changed, is limited to the same "category type". Applications components can only be changed into other applications, Data components can only be changed into other data components, and Constructs can only be changed into other constructs.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -4524,9 +4524,14 @@ export class Eagle {
         let eligibleComponents : Node[];
 
         if (this.selectedNode().isData()){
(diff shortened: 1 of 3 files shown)


M-Wicenec (changes requested):
Worked as intended for me. Just that small change we talked about for the edit category selection menu to have the currently used category pre-selected

PR #416, opened 2022-07-07
Delete bug on ports in the argument table

M-Wicenec:


M-Wicenec:
the removePortFromNodeByIndex function wasnt working for ports in the arguments table as it references the current port's index from the getInput/OutputPorts() function. this doesnt work when the index provided is in relation to all fields on a node.
since at the points where we are calling the function we are already holding the field, i am now instead passing the fieldId and using the getPortIndexById() function to get what we want. 
if also added a few ko statements to add different delete buttons with seperate functions for output ports, input ports and a shared one for component parameters and application arguments

james-strauss-uwa (approved):

PR #424, opened 2022-07-15
Eagle 814

james-strauss-uwa:
Fixing two bugs:
1. When adding a new application argument, via the '+' button and 'custom' option. The new field would have an id equal to the empty string.
2. An edge could be dragged to a port with an empty id. If multiple ports on the node had empty ids, then the renderer could not determine which was the correct port, and would just choose the first.

The solution to part 1 was to add the id when creating a new application argument. The code here has a bit of technical debt, so I added some TODO comments. Also, I added a Eagle.printNodeFieldsTable() function to easily print the status of all fields on a node.

For part 2, I modified the edge drag auto-complete so that it would not suggest ports with empty ids. Also, modified the Edge.isValid function to judge edges between ports with no ids as invalid. So normal users will not be able to draw them. Expert users (with "Allow invalid edges" setting set to true) will still be able to draw those edges. Finally, I modified Utils.checkGraph() to produce an error if a field has no id

I've checked that this doesn't break:
- adding fields from the parameter table
- editing existing fields from the inspector + modal
- adding component parameters from the inspector

It seems to be OK, but if you could take a look at those things and anything else you can think of, that would be a great help. Thanks.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -356,7 +356,7 @@ export class Eagle {
         console.log('srcNode', p1x, p1y, 'destNode',p2x, p2y)
         $('#nodeList .col').append('<div class="positionPointer" style="width:5px; height:5px;position:absolute;background-color:red;z-index:1000000;top:'+p1y+'px;left:'+p1x+'px;"></div>')
         $('#nodeList .col').append('<div class="positionPointer" style="width:5px; height:5px;position:absolute;background-color:blue;z-index:1000000;top:'+p2y+'px;left:'+p2x+'px;"></div>')
-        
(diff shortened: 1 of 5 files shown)


awicenec (commented):
The changes are working fine. The only thing I would include is to move the 'Custom' option to the start of the pull-down list, because that's the most often used option. I think there is a separate ticket for that, though.

awicenec:
One more thing, when you merge, please up the release number to 4.5.3

james-strauss-uwa:
OK, merged and I have tagged the new version as v4.5.3

PR #426, opened 2022-07-19
Eagle 737

james-strauss-uwa:
Added '+' buttons to the "input ports" and "output ports" sections of the inspector


M-Wicenec (approved):
Looks good. The ticket also mentions moving the custom option to the top of the selection drop-down. If youd like to do that in this ticket, that would be one change to consider.

PR #429, opened 2022-07-28
Eagle 767

james-strauss-uwa:
Changes to Field data types, and errors.

Errors used to be stored as just a string, but now I've added Errors.Issue, which stores:
- the error message
- a 'show' function which will select the object, so the user can investigate
- a 'fix' function, if some obvious change will fix the issue
- a 'fixDescription', a string that lets the user know what action the fix function will perform

Buttons to 'show' errors, or 'fix' errors should appear wherever the errors are shown. In the errorsModal, and now in the Node/Edge inspectors.

The errors modal now has a "Fix All" which just runs all the fixes in the errors/warnings list. Note that sometimes fixing an error will make another one appear. For example, if a port has an unknown data type, a fix function might set the datatype to be the same as another port to which it's connected. However, if the port changes to something like an Integer, it may then trigger a warning if it does not have a default value. So sometimes we have to click "fix all" twice.

Fixed a bunch of other small related bugs:
- fixed the "isEvent" toggle (the clock icon) next to ports in the inspector
- fixed the edge change data type function, accessible from the pencil icon in the edge inspector
- moved from using the 'Complex' type, to using 'Object'
- minor fixes to undo/redo

It now displays node-related errors and warnings in the Node inspector. Similar to the way the Edge inspector already did it.

The unit test failure is due to adding a 'event' attribute to fields in the fields list. And that attribute is not in the version of the graph in EAGLE-graph-repo, so the JSON doesn't match. I'll update the version of the graph in EAGLE-graph-repo once this code is merged.


james-strauss-uwa:
Just spoke to @M-Wicenec about this, here is the feedback:

- After changing the type of a port in the parameters table, checkGraph is not re-run, so the number of errors does not change.
- We should add some feedback after "Fix All". Perhaps append a message to the bottom of the errorsModal with information on the number of items fixed, and the number that could not be fixed.
- Every error and warning in the lists should at least have a "Show" button, even if no "Fix" is possible.
- We could make the "Fix All" button loop over the fixes multiple times. So that it also performs any fixes that come up as a result of the first set of fixes.

PR #441, opened 2022-09-07
Eslint aug 2022

james-strauss-uwa:
Initial aims:

- to reduce the size of Eagle.ts
- run linter (eslint) on all code

The size of Eagle.ts size has changed from ~5500 lines to ~4100 lines. Largely by moving code out into other files, and the creation of a few new files.

Moved Category and CategoryData out of Eagle.ts.

Moved some constants to Config.ts

Made many of the other files available in the global namespace. For example now: Eagle, Utils, Config, Category, Errors, Hierarchy, Setting, Repositories, ParameterTable, SideWindow are all available globally.

Also added Eagle.getInstance(), a static function on Eagle, that allows us to get the "eagle" object from anywhere in the code (where Eagle is imported).

All occurrences of "var" have been changes to "const" or "let".


M-Wicenec (approved):
Looks good, i had a skim over the file changes and a fairly thorough testing of eagle. I didnt run into any bugs

PR #450, opened 2022-09-20
Liu 303 - Correcting Scheduler Algorithms

pritchardn:
This sibing PR of ICRAR/daliuge/pull/202, removes the simple and none translation options and adds No. of Nodes to the remaining options.
The default parameters provided should successfully translate a graph for all options.


src/Config.ts
```diff
@@ -36,8 +36,7 @@ export class Config {
         "metis",
         "mysarkar",
         "min_num_parts",
-        "pso",
-        "simple"
+        "pso"
```
james-strauss-uwa, line 39: It seems weird that we just remove one element here, but I can see that removing the first element will require us to re-index all the genPGT() calls in the html. So it makes sense to leave element 0 in place.

Maybe just add a comment that element 0 ("none") is not used.

OR maybe we should just remove the whole "translationAlgorithms" variable, and move the strings ("metis", "mysarkar" etc) directly into the genPGT() calls, changing the first argument to genPGT() from algorithmIndex:number to algorithmName:string?

I don't think we use the translationAlgorithms variable anywhere else, so maybe we can just remove one layer of unnecessary indirection.

Up to you

james-strauss-uwa (approved):
Looks good.

src/Config.ts
pritchardn: I like the idea of changing the variable from an index to the string directly so I'll go ahead and do that.
There is some method to only removing a single element - in the API 'none' is a valid 'algorithm' choice, it just doesn't work at the moment. Simple on the other hand, doesn't exist at all.

PR #456, opened 2022-10-04
Eagle 841

M-Wicenec:
Key parameter flag and table added


awicenec:
A few points to address before the merge:

1. Re-word the hover of the global table to "Configure Key Graph Attributes"
2. Change header of the key parameter table to "Key Graph Attributes" (this is then also consistent with the hover text over the heart in the node attributes table)
3. Since these are by definition are all key parameters, maybe the heart column in the Key Graph Attributes Table should not be shown?

Future considerations (new tickets, roughly in priority order).

1. Highlight node in graph when selecting a node row in the Key Graph Attributes Table. Once that is done we can drop the Key column, which is not that useful, since the keys of all the nodes is not visible on the graph.
2. Show heart also in inspector side panel (as toggle button?)
3. Add a toggle button to the normal node attributes table header to allow showing all attributes, independent of the "Show only key graph attributes" setting (see below). This toggle button should not be visible in minimal mode (at least).
4. Change header of Edit modal to show the name and a heart next to it, if it is a key parameter.
5. Change text and function of the current setting "Show DALiuGE runtime parameters" to "Show only key graph attributes". This would both invert and extend the current setting, i.e. if set only the key attributes would be shown and not only for the component parameters.
6. Allow re-sizing table modals, in particular changing the height would be useful
7. Allow collapsing Key Graph Attributes Table to show only a head row for each node in order to make it easier to deal with big graphs and lots of key parameters.

PR #470, opened 2022-10-21
Eagle 838

james-strauss-uwa:
This separates the "useAs" attribute of fields into two: "parameterType" and "usage"

parameterType can be: ComponentParameter, or ApplicationArgument
usage can be: NoPort, InputPort, OutputPort, InputOutput

It makes changes to the parameterTable and the editField modal. The merge was difficult, so please check that the recently added functionality is still present.

The other big change is the removal of the "inputPorts" and "outputPorts" arrays on the Node JSON. Instead, everything is in a single "fields" array. This is quite a big change, so needs to be coordinated with DALiuGE

The "remote" load/save test is still failing, but I'll update that at the last moment, otherwise it will fail for other branches.


M-Wicenec (approved):
looks good, i just added some css for the new columns to set their width

james-strauss-uwa:
@awicenec This code would change the output graph JSON (for each Node) from:
- a "fields" array for "component parameters" and "application arguments"
- a "inputPorts" array containing input ports
- a "outputPorts" array containing output ports

to :
- a "fields" array containing "component parameters", "application arguments", "input ports" and "output ports", where each element has "parameterType" and "usage" attributes

Is it OK to make this change to the EAGLE output format, or should we wait?

PR #523, opened 2023-02-01
Eagle 838 2

james-strauss-uwa:


awicenec:
I have now merged my daliuge branch and that should enable the usage of the new format as well. Still need to do a release on PyPi and dockerHub

james-strauss-uwa:
OK, great.

I added a setting (in the developer tab) called "Use Old Output Format". It is true by default. If you change this to false, you can test output with the new format.

We can remove that setting once we move to the new format, which sounds like it could be soon.

PR #536, opened 2023-02-21
Eagle 969 - move settings and tooltip set up from eagle into their respective files

james-strauss-uwa:
Moved settings and keyboardshortcuts out of Eagle.ts

No longer using knockout ObservableArray for Eagle.shortcuts. I guess this could be a likely source of errors, but it seems to work for me.

No longer using strings in Utils as the keys for Settings in localStorage, moved the strings to Settings.

Moved several settings-related enums from Eagle to Settings. UIMode etc.

Just FYI to see where everything has moved. If you have time, maybe check the way the '3' and '4' key shortcuts change when an object is selected vs. nothing is selected. Seems to work for me.


M-Wicenec (approved):
looks good, did test and play around with it a bit as well, didnt encounter any errors

PR #538, opened 2023-03-09
Eagle 950 3

M-Wicenec:


src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+        cooldown = true
+        setTimeout(function(){
+            cooldown = false
+        }, 500)
+    }
+
+    initiateTutStep = (direction:string) :void => {
+        const eagle = Eagle.getInstance()
+
+        var tutStep = activeTut.getTutorialSteps()[activeTutCurrentStep-1]
```
james-strauss-uwa, line 96: Why -1 here?

Use const

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+
+        //if there is a preFunction set, then we execute it here
+        var preFunction
```
james-strauss-uwa, line 99: Use let

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+export namespace TutorialStep {
+    export enum Type {
+        Info,
+        Press,
+        Input,
+        Condition
+    }
+}
+
+export namespace Wait {
```
james-strauss-uwa, line 425: Do you need another namespace here? I think this could belong in the TutorialStep namespace.

TutorialStep.Wait.Modal etc.

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+            preFunction = tutStep.getPreFunct()
+        }else if(direction === 'prev'){
+            preFunction = tutStep.getBackPreFunct()
+        }
+        
+        if(preFunction != null){
+            preFunction(eagle)
+        }
+        
+        //we always pass through the wait funciton, it is decided there if we actully wait or not
```
james-strauss-uwa, line 111: Minor spelling: "function", "actually"

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+        this.initiateWaitForElement(tutStep.getWaitType())
+    }
+
+    initiateWaitForElement = (waitType:Wait.Type) :void => {
```
james-strauss-uwa, line 115: This function seems to do more than wait. It calls pickStepType(), which then initiates/runs the step.

It seems like this behaviour is a bit hidden behind the name "wait"

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+                alternateHighlightSelector = selectorElement
+            }else{
+                selectorElement = selectorElement()
+            }
+            elementAvailable = selectorElement.hasClass('show')
+        }else if (waitType === Wait.Type.Element){
+            if(selectorElement().length){
+                elementAvailable = true
+            }else{
+                return
```
james-strauss-uwa, line 160: I think this case should add a warning/error to the console.

If were waiting for an element, and zero elements were found using the selector, then we should report an error.

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+                    console.warn('waiting for next tutorial step element timed out')
+                }
+            }, 2000)
+        }
+    }
+
+    waitForElement = (waitType:Wait.Type) :void => {
+        var tutStep = activeTut.getTutorialSteps()[activeTutCurrentStep-1]
+        var elementAvailable = false
+        var selectorElement = tutStep.getSelector()
```
james-strauss-uwa, line 134: Use of selectorElement is a little confusing here. I would expect tutStep.getSelector() to return a string. But the result is called 'selectorElement', so I think it is a HTML element. But within this function, selectorElement is sometimes called like a function, and sometimes not.

Declare the type of the variable. Use let?

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+        if(elementAvailable){
+            this.pickStepType(tutStep, alternateHighlightSelector)
+            clearTimeout(waitForElementTimer);
+            waitForElementTimer = null;
+        }else{
+            return
+        }
+    }
+
+    pickStepType = (tutStep:TutorialStep, alternateHighlightSelector:JQuery<HTMLElement>) :void => {
```
james-strauss-uwa, line 176: This function, pickStepType, calls initiateXStep(), where X is the Type of the step. Would a better name be initiateStep() (instead of pickStepType)

alternateHighlightSelector could be the wrong name here, it seems to be an element?

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+    //these are ground work for fufture tutorial system functionality
+    initiateInputStep = (tutStep:TutorialStep,alternateHighlightSelector:JQuery<HTMLElement>) : void => {
+        console.log('initiating input step')
+    }
+
+    initiateConditionStep = (tutStep:TutorialStep, condition:string,alternateHighlightSelector:JQuery<HTMLElement>) : void => {
+        console.log('initiating condition step')
+    }
+
+    highlightStepTarget = (selector:JQuery<HTMLElement>) : void => {
```
james-strauss-uwa, line 228: Maybe 'selector' parameter is actually an element

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+        
+        //in order to darken the screen save the selection target, we must add divs on each side of the element.
+        var coords = selector.offset()
```
james-strauss-uwa, line 231: Use const/let here

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+        $('body').append("<div class='tutorialHighlight' style='top:"+top_actual+"px; right:0px;bottom:"+bottom+"px;left:"+right+"px;'></div>")
+        //bottom
+        $('body').append("<div class='tutorialHighlight' style='top:"+bottom_actual+"px; right:0px;bottom:0px;left:0px;'></div>")
+        //left
+        $('body').append("<div class='tutorialHighlight' style='top:"+top_actual+"px; right:"+left+"px;bottom:"+bottom+"px;left:0px;'></div>")
+    }   
+
+    openInfoPopUp = () :void => {
+
+        var step = activeTut.getTutorialSteps()[activeTutCurrentStep-1]
```
james-strauss-uwa, line 253: Why -1 here?

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+    }
+}
+
+export class TutorialStep {
+    private title : string;
+    private text : string;
+    private type : TutorialStep.Type;
+    private waitType : Wait.Type;
+    private selector : () => void;
+    private preFunction : (eagle:Eagle) => void;
```
james-strauss-uwa, line 373: Some comments of the purpose of/requirement for the forward and back funcs would be good.

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+    private backPreFunction : (eagle:Eagle) => void;
+
+    constructor(title : string, text : string, type : TutorialStep.Type,waitType: Wait.Type, selector:() => void, preFunction:(eagle:Eagle) => void, backPreFunction:(eagle:Eagle) => void){
```
james-strauss-uwa, line 376: I would expect the selector to be a string-type here

james-strauss-uwa (changes requested):

src/Tutorial.ts
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    getDescription = () : string => {
+        return this.description;
+    }
+
+    initiateTutorial = (tutorialName:string) : void => {
+        const that = this;
+
+        Eagle.tutorials.forEach(function(tut){
```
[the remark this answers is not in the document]
M-Wicenec: good point, I alternatively just removed the 'that = this' line by calling the function using the activeTut global variable instead.

james-strauss-uwa (approved):

PR #596, opened 2023-05-31
Eagle 970

M-Wicenec:


src/Node.ts
```diff
@@ -573,6 +573,7 @@ export class Node {
     }
 
     setCategory = (category: Category): void => {
+        console.log('Node.setCategory()', category)
```
james-strauss-uwa, line 576: Can remove console.log

james-strauss-uwa (approved):
Looks good, just one console.log that can be removed. I can't think of anything else to remove.

PR #616, opened 2023-08-01
Eagle 1039

james-strauss-uwa:
Could you take a quick look at this. Sorry, its a large code change.

The major changes are to use a common interface (a modal previously used for displaying/fixing graph errors) for three purposes:

1. Displaying/Fixing graph errors
2. Displaying/Fixing errors during loading of palettes and graphs
3. Displaying/Performing updates to components based on newer palette definitions

It's useful to enable the following settings:

- Show edge/node errors/warnings in inspector: Warnings
- Show File Loading Warnings: Enabled

And test:

- Loading graphs and palettes, clicking "Show", "Fix" buttons on single issues, and "FixAll" button on the modal
- Fixing a graph with errors (again use "Show", "Fix" and "Fix All")
- Updating a graph (for example the graph in the EAGLE repo in tests/data/component-update-test.graph)

Also please check that removing line 345 from src/Setting.ts is OK


M-Wicenec:
Removing that setting is fine, that was my mistake

the fix all and individual fixing of issues in the modal on load are working great.

Problems:
the individual fixing of issues in the inspector was not doing anything for me, both for nodes in the palette and nodes in the graph

there is no way to get back to the version of the error modal that displays both, errors in the graph and in the palettes.
maybe we should have both available in the normal error modal with tabs to switch between graph and palette at the top?

PR #640, opened 2023-10-24
Eagle 1089

james-strauss-uwa:
Draw edge while dragging from a port. If edge is dragged into open space, edge will persist until the context menu is closed.

Still only works at scale == 1

Had to make the "mouse position" and the "boolean for dragging edge drawing" observables


james-strauss-uwa:
I've added the "auto complete suggested edge" drawing now, so this can be re-reviewed

PR #647, opened 2024-03-19
Make it hard to save a graph with no name (or a bad name)

james-strauss-uwa:
When adding any node to a graph, check that the graph is named. If not, request a name from the user.
When saving a graph locally, check that the graph is named. If not, abort and notify user.


M-Wicenec (approved):
code looks good, works well.

PR #650, opened 2024-03-21
Factored out some common validation code into a single function within Utils

james-strauss-uwa:


M-Wicenec (approved):
good idea! still working as before.

PR #652, opened 2024-03-22
Add new 'Impossible' value to the Eagle.LinkValid enum

james-strauss-uwa:
These are a class of edges that can never be created, even if the user has set 'allow invalid edges' to true.


M-Wicenec (approved):
👍

PR #664, opened 2024-04-24
Linting apr 2024

james-strauss-uwa:
- General cleanup
- Switched declaration of static functions to a different syntax (as recommended by a linter)
- Use const/let instead of var
- Use for ... of loops where appropriate
- Delete unnecessary imports
- Sort some imports alphabetically
- Removed some commented code
- Fixed some spelling mistakes
- Standardised indenting
- Replaced the "any" type with specific types where possible
- Added types (MouseEvent/KeyboardEvent etc) where possible
- Replaces `(<any>window).eagle` with `Eagle.getInstance()`
- Replaced deprecated jquery functions bind/unbind/click/focus/blur with on/off/trigger
- Replaced deprecated event.which with event.key
- Where functions return their own class (e.g. TutorialStep) use "this" keyword to allow subclassing
- Removed some function parameters where they were unused
- Use nullish coalescence operator (??) where possible when reading possibly undefined values


M-Wicenec (approved):
changed the one TODO thing we talked about. otherwise looks good

PR #665, opened 2024-04-26
Eagle 1198

M-Wicenec:


src/GraphRenderer.ts
```diff
@@ -83,29 +83,11 @@ ko.bindingHandlers.nodeRenderHandler = {
                 const eagle : Eagle = Eagle.getInstance();
                 node = eagle.logicalGraph().findNodeByKey(node.getParentKey())
             }
-            GraphRenderer.resizeConstruct(node,false)
+                GraphRenderer.resizeConstruct(node,false)
```
james-strauss-uwa, line 86: Is this indenting correct?

src/LogicalGraph.ts
```diff
@@ -682,6 +707,7 @@ export class LogicalGraph {
                 maxDepthOverlap = overlap;
             }
         }
+        // console.log('FINAL node at location: ', maxDepthOverlap.getName())
```
james-strauss-uwa, line 710: maybe remove this

src/LogicalGraph.ts
```diff
@@ -650,30 +650,55 @@ export class LogicalGraph {
...
+                    overlaps.push(node);
+                }
+            }   
         }
 
         // once found all the overlaps, we return the most-leaf (highest depth) node
         let maxDepth: number = -1;
         let maxDepthOverlap: Node = null;
 
+        // console.log(overlaps)
```
james-strauss-uwa, line 700: maybe remove this

src/GraphRenderer.ts
```diff
@@ -1235,8 +1229,10 @@ export class GraphRenderer {
 
         GraphRenderer.dragSelectionHandled(true)
         eagle.isDragging(false);
-        eagle.draggingNode(null)
+        eagle.draggingNode(null);
         
+        //this is a little hack to force affected constructs to update their size on mouse up its neccessary
+        eagle.selectedObjects(eagle.selectedObjects())
```
james-strauss-uwa, line 1235: Could we use eagle.selectedObjects.valueHasMutated() here?

src/Eagle.ts
```diff
@@ -763,6 +763,47 @@ export class Eagle {
         return false;
     }
 
+    getOutermostSelectedNodes = () : Node[] => {
+        const outermostNodes : Node[] = []
+        const selectedNodes = this.selectedObjects()
+        const eagle = this
+
+        selectedNodes.forEach(function(object){
+            if(object instanceof Node){
```
james-strauss-uwa, line 772: Maybe rather than nest most of the function inside this if, you could invert it as an early exit.

if (!(object instanceof Node)){
    return outermostNodes;
}

then continue with the rest of the function

james-strauss-uwa (approved):
Just tiny comments, otherwise looks good.

src/GraphRenderer.ts
M-Wicenec: it is now :)

src/GraphRenderer.ts
M-Wicenec: yep that also works, good call

PR #666, opened 2024-04-29
Eagle 545

james-strauss-uwa:
Makes loading of local files completely client-side. No round-trip to the server is required.

Also removed the server end-point /uploadFile, since it is no longer used.

To test:
- load a local graph
- insert a local graph
- load a local palette


M-Wicenec (approved):
looks good, side note. saving palettes to local seems to be bugged

PR #667, opened 2024-04-30
Eagle 1197

james-strauss-uwa:
Allow EAGLE to function offline.

Previously EAGLE would always try to fetch the 'Builtin Components' and 'Component Templates' palettes, and the graph schema definition file. If offline, these fetches would fail, leaving EAGLE in a semi-broken state.

Now:

1. If EAGLE is online. It runs as usual, but puts a copy of the palettes and schema in HTML localStorage.
2. If EAGLE is offline, and starting for the first time, then we can't fetch the palettes. So we just do the best we can with no palettes/schema, but at least EAGLE is functional.
3. If EAGLE is offline, and has been online in the past, then we load the palettes and schema from HTML localStorage.


james-strauss-uwa:
Updated this branch to avoid using navigator.onLine, since this was not accurate on at least one test machine.

Now the code attempts to fetch the palettes/schema and if the fetch fails, it will attempt to load the palettes/schemas from localStorage.

Also, those requests to the eagleServer got INTERNAL SERVER ERROR 500 as a response, because the request would trigger an uncaught exception. Now the exceptions are caught on the server, and the error message is sent to the client for display.

We also spoke about a online/offline indicator, perhaps near to the EAGLE logo. But I'm going to split that out into another ticket.

Could you take another look. Maybe on your home PC as well, when you have a chance.

PR #668, opened 2024-04-30
Eagle 1200

james-strauss-uwa:
Fixed bug where dropping a new node (B) on top of an existing node (A), would make A the parent of B, even if A was not a group node type. Resulting in an invalid parent relationship.


M-Wicenec (approved):
issue fixed, thanks

PR #675, opened 2024-05-13
Added 'browse dockerhub' button to node inspector, for docker nodes only

james-strauss-uwa:
Uses existing button handler code that was previously unused.

Fixed bug where EAGLE thought the names of the tag and digest field in a Docker component were called "tag" and "digest" instead of "docker_tag" and "docker_digest".

Updated query to dockerhub where API had changed slightly ("user" vs. "namespace")

Added new red notification for one error case, where no tags are available for the requested image.
Added new green notification for success.

### DIFF
--- a/src/Daliuge.ts
+++ b/src/Daliuge.ts
@@ -61,8 +61,8 @@ export namespace Daliuge {
 
         // docker
         IMAGE = "image",
-        TAG = "tag",
-        DIGEST = "digest",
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
looks good. ive added the button in the parameter table for docker nodes as well, i made a little helper function for it that first closes the table then calls your function.

PR #676, opened 2024-05-13
Removed the dataType attribute from Edge

james-strauss-uwa:
Removed the 'dataType' attribute from Edges in the graph. Since we already have the type defined in the source and destination port of the edge, storing the type in the edge too is just redundant, and increases the likelihood of mismatches.

There is also a related merge on DALiuGE to handle the changes on that end.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -2629,7 +2629,7 @@ export class Eagle {
         }
 
         // if input edge is null, then we are creating a new edge here, so initialise it with some default values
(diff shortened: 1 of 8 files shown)


james-strauss-uwa:
Actually, Andreas has merged the DALiuGE part of this into master. So I'd like to try to get this into master too before we do a release.

Like I said, I think it won't affect to much of Eagle, since the dataTypes were not very important. But it does touch:
- drawing edges
- edge editing modal
- graph checking
- edge inspector UI

If you could have a quick play around with these to test things, that would be great. No rush.

M-Wicenec (approved):
code looks good. i havent found any new bugs yet

PR #679, opened 2024-05-14
Modified insertGraph function to update node subjectKey attribute usi…

james-strauss-uwa:
Given a graph with two nodes, a File and a (connected) Comment. This seems to work when:
- Duplicating just the comment. The new comment is connected to the old File.
- Duplicating both nodes. The new comment is connected to the new File.
- Saving the graph to git, then inserting from git. The new comment is connected to the new File, both inside a Subgraph.


M-Wicenec (approved):
This is working great!

PR #681, opened 2024-05-16
Eagle 1050 - Open/close builtin palettes when the OPEN_DEFAULT_PALETTE setting is toggled.

james-strauss-uwa:
Loads or removes the two builtin palettes when the user toggles the "Open default palettes on startup" setting in the "User Options" tab.

Previously each Setting object had a "buttonFunc" attribute. This function would be called when the setting button was clicked (if the Setting was a Button type). Non-Button type settings would not use the buttonFunc attribute.

I modified buttonFunc to eventFunc, and called it on Boolean-type settings as well as Button-type settings. So when the checkbox is toggled, the eventFunc is called.

Also split the "loadDefaultPalettes" functionality out of Eagle into its own function, so that it could be called by the eventFunc attached to the OPEN_DEFAULT_PALETTE setting.

If you could test:
- toggling the setting
- starting Eagle with both values
- switching UI modes

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -2231,7 +2231,19 @@ export class Eagle {
     // if currently shown, just remove them from the palettes list
     // if currently not shown, fetch them from the remove source and add to palettes list
     toggleDefaultPalettes = () : void => {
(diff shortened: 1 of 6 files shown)


M-Wicenec (commented):
toggling the button and loading works in both states.
changing ui mode however, does not immediately update the ui, it is correct after reloading eagle

PR #687, opened 2024-05-24
Eagle 1225 - Enable Input and Output Application for subgraph construct

james-strauss-uwa:
Modified CategoryData to allow input and output applications on SubGraph nodes

There were a few of places where we insert graphs:

- Create subgraph from selection
- Insert remote graph
- Insert local graph

So I factored out some of the common code and placed it in two new Utils functions

- createSubgraphParent
- duplicateNode

NOTE: there may be some other places in EAGLE where we can use Utils.duplicateNode. Instead of node.clone() and then manually updating 1) the node key of the node and embedded applications 2) the ids of all the fields in the node and embedded applications


james-strauss-uwa:
@M-Wicenec  this one is ready to go now. I've added everything I intended to.

M-Wicenec (approved):
works and code looks good

PR #697, opened 2024-06-13
EAGLE-1231: Added 'encoding' attribute to fields

james-strauss-uwa:
Added an Enum to Daliuge class for all possible encoding values.
Updated ParameterTable with an additional column that displays the encoding values.

NOTE: The 'encoding' column in the table seems to default to 'not shown'. The user has to manually enable the column in the column visibilities.

## Summary by Sourcery

This pull request adds an 'encoding' attribute to the Field class and updates the ParameterTable to include a new column for encoding values. The encoding column is not shown by default and must be manually enabled by the user.

* **New Features**:
    - Introduced an 'encoding' attribute to the Field class to specify data encoding types.
* **Enhancements**:
    - Updated the ParameterTable to include an additional column for displaying encoding values, which can be toggled by the user.

### DIFF
--- a/src/Daliuge.ts
+++ b/src/Daliuge.ts
@@ -85,6 +85,15 @@ export namespace Daliuge {
         Python = "Python"
     }
 
+    export enum Encoding {
+        Binary = "binary",
+        Pickle = "pickle",
+        Dill = "dill",
+        Npy = "npy",
+        Base64 = "base64",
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
looks good, ive fixed the defaults for the new column visibilities. also changed the load from local storage function to use defaults when a value for the newly added column isnt set

PR #702, opened 2024-06-20
Eagle 1248 - Old graph (cont_img_mvp.graph) with embedded app inside Service breaks EAGLE

james-strauss-uwa:
The graph JSON indicates that several "Service" nodes, for example "GSM", should have output ports. EAGLE would try to add the output port during load. But according to CategoryData, Service nodes can't have any outputs. So EAGLE would assume that the port belonged on an embedded app within the Service.

The function that adds the port to an embedded app would first check if the embedded app existed. If not, an embedded app would be created. This caused all sorts of mess.

Now EAGLE will not automatically add embedded nodes to non-Constructs.

I also took the opportunity to add several "Fix" functions for errors on that graph:

- Source port doesn't exist on source node
- Destination port doesn't exist on destination node
- Source port is not an output port
- Destination port is not an input port

And improved the fix function for this error
- Source and destination ports don't match

Also, a couple of the fix functions could cause issues if they attempted a fix that was already done. So I've added checks to a couple of the fix functions where they abort the fix if it is no longer required.

## Summary by Sourcery

This pull request addresses a critical bug where EAGLE would incorrectly add output ports to Service nodes, causing issues with embedded applications. It also introduces several enhancements to error handling and fix functions for port-related errors, ensuring more robust and accurate graph validation.

- **Bug Fixes**:
    - Fixed issue where EAGLE would incorrectly add output ports to Service nodes by preventing automatic addition of embedded nodes to non-Constructs.
    - Added checks to prevent fix functions from attempting redundant fixes.
- **Enhancements**:
    - Improved error handling and added fix functions for various port-related errors, including missing source/destination ports and incorrect port types.
    - Enhanced the fix function for mismatched source and destination ports to provide more detailed error messages.

### DIFF
--- a/src/Edge.ts
+++ b/src/Edge.ts
@@ -354,13 +354,15 @@ export class Edge {
 
         // check if source port was found
         if (sourcePort === null) {
-            Edge.isValidLog(edgeId, Eagle.LinkValid.Impossible, Errors.Show("Source port doesn't exist on source node", function(){Utils.showEdge(eagle, edgeId);}), showNotification, showConsole, errorsWarnings);
+            const issue: Errors.Issue = Errors.ShowFix("Source port (" + sourcePortId + ") doesn't exist on source node (" + sourceNode.getName() + ")", function(){Utils.showEdge(eagle, edgeId)}, function(){Utils.addSourcePortToSourceNode(eagle, edgeId)}, "Add source port to source node");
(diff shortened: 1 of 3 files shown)


M-Wicenec (approved):
nice, no more errors, new fixes work great. we may get some merge conflicts with the graph errors branch and this one.

PR #704, opened 2024-06-21
EAGLE-1247: Problem when saving graph without ".graph" extension

james-strauss-uwa:
Adds client-side validation of the graph or palette filename. Uses bootstrap 'is-valid/is-invalid' CSS to display validity.

Note: the "git commit" modal does not actually prevent the user from proceeding with a invalid filename. It just warns the user. However, once the modal is dismissed, if the filename does not end with .graph or .palette as appropriate, that extension will be added to the filename.

To test:
- Open a graph, commit to git using "save as"
- Open a graph, use the navbar button to create new palette from graph nodes, then click the "three dot" menu in the palette header, choose "save as"

## Summary by Sourcery

This pull request addresses the issue of saving graph and palette files without the correct extensions by adding client-side validation. The 'git commit' modal now provides visual feedback on filename validity and automatically appends the correct extension if missing.

- **Bug Fixes**:
    - Added client-side validation to ensure graph and palette filenames end with '.graph' and '.palette' respectively.
- **Enhancements**:
    - Updated the 'git commit' modal to display validation feedback using Bootstrap's 'is-valid/is-invalid' CSS classes.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -1700,7 +1700,7 @@ export class Eagle {
             }
         }
 
-        Utils.requestUserGitCommit(defaultRepository, Repositories.getList(defaultRepository.service), fileInfo().path, fileInfo().name, (completed : boolean, repositoryService : Repository.Service, repositoryName : string, repositoryBranch : string, filePath : string, fileName : string, commitMessage : string) : void => {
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
works great, although, I would have preferred to automatically resolve the extension problem by replacing invalid extensions or adding the correct one to save the user the hassle, but this also fixes the issue :)

PR #706, opened 2024-06-25
EAGLE-1253: Only show Parameter Table encoding drop-down for fields being used as ports.

james-strauss-uwa:
Only show Parameter Table encoding column by default in expert mode.
Default encoding is now Pickle instead of UTF8.

## Summary by Sourcery

Updated the default encoding to Pickle instead of UTF8 and modified the Parameter Table to only show the encoding drop-down for fields being used as ports.

- **Enhancements**:
    - Parameter Table encoding column is now only shown for fields being used as ports.

### DIFF
--- a/src/Field.ts
+++ b/src/Field.ts
@@ -54,7 +54,7 @@ export class Field {
         this.options = ko.observableArray(options);
         this.positional = ko.observable(positional);
         this.keyAttribute = ko.observable(keyAttribute);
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
looks good, 
I adjusted the colour of N/A encoding cells to be greyed out to reduce visual clutter and show that there are no changes that can be made there

PR #709, opened 2024-06-26
EAGLE-1218: Specify height of the git commit modal

james-strauss-uwa:
Just wanted to check with you that this is the best way to fix the problem

## Summary by Sourcery

This pull request enhances the user interface by setting a fixed height for the git commit modal, ensuring consistent display across different scenarios.

- **Enhancements**:
    - Added a fixed height of 420px to the git commit modal for better UI consistency.

### DIFF
--- a/static/base.css
+++ b/static/base.css
@@ -1424,9 +1424,12 @@ select.form-control{
     height: 100%;
 }
 
-
 #inputTextModalInput {
     font-family: monospace;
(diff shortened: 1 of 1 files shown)


M-Wicenec (approved):
I think this is reasonable, since we know the content is static and wont change in height, fixing it is fine.

PR #711, opened 2024-06-26
Eagle 1244

M-Wicenec:
keyboard shortcuts and the tab change buttons in the right window but used different methods of changing the right window mode. this meant that the behaviour was different depending on which action you chose to switch the right window mode. 
I've simply created a helper function to call in both cases to keep this process more simplified.

## Summary by Sourcery

This pull request refactors the way the right window mode is changed by introducing a helper function `changeRightWindowMode`. This ensures consistent behavior across different methods of changing the right window mode, such as keyboard shortcuts and tab change buttons.

- **Enhancements**:
    - Unified the method for changing the right window mode by introducing a helper function `changeRightWindowMode`.


src/Eagle.ts
```diff
@@ -730,6 +730,18 @@ export class Eagle {
         this.rightWindow().mode(Eagle.RightWindowMode.Hierarchy)
     }
 
+    changeRightWindowMode(requestedMode:Eagle.RightWindowMode) : void {
+        this.rightWindow().mode(requestedMode)
+        this.rightWindow().shown(true); 
+
+        //if we are intentionally switching to the hierary, then we do not want the right window mode to switch to the inspector when we change node selection.
+        if(requestedMode===Eagle.RightWindowMode.Hierarchy){
```
james-strauss-uwa, line 738: This could be one line:

this.hierarchyMode(requestedMode === Eagle.RightWindowMode.Hierarchy);

james-strauss-uwa (approved):

PR #712, opened 2024-06-26
EAGLE-1098: Fixed issue where selectAll then delete would not do anything

james-strauss-uwa:
The main issue was that "selectAllFromGraph" (Ctrl-A) did not set Eagle.selectedLocation(). So deleteSelection() not take the code path for either the graph or palette location, it would just exit without doing anything.

So the main fix here was just a one liner on Eagle.ts line 3153.

However, I did a bit of refactoring at the same time.

- Warn if location is unknown
- Removed string-based mode arguments to delete functions. The mode was used to determine the location, so just passed the location instead.

## Summary by Sourcery

This pull request fixes a bug where selecting all items and then deleting them would not work. It also includes refactoring to improve the delete functions by removing string-based mode arguments and using location directly, and adds a warning for unknown locations during deletion.

- **Bug Fixes**:
    - Fixed issue where 'selectAll' followed by 'delete' would not perform any action by ensuring 'selectAllInGraph' sets the selected location.
- **Enhancements**:
    - Refactored delete functions to remove string-based mode arguments and use location directly.
    - Added warning for unknown location during deletion.


M-Wicenec (approved):
great catch, and a lot of nice code improvements

PR #714, opened 2024-06-27
EAGLE-1256: Re-use Edge.isValid within the edge auto-complete system

james-strauss-uwa:
The edge auto-complete system had a seperate and not-quite-as-good implementation of checking for suitable edges. I've updated it to re-use the Edge.isValid function.

This allowed findMatchingNodes() to be removed, and findMatchingPorts() could be simplified.

EAGLE remains performant, because the check for suitable destination ports only happens once, when dragging from an edge first starts.

## Summary by Sourcery

This pull request refactors the edge auto-complete system to reuse the Edge.isValid function, leading to the removal of the findMatchingNodes function and simplification of the findMatchingPorts function. These changes enhance the consistency and maintainability of the codebase.

- **Enhancements**:
    - Refactored the edge auto-complete system to reuse the Edge.isValid function, improving consistency and maintainability.
    - Simplified the findMatchingPorts function by removing redundant logic and directly using Edge.isValid for validation.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -2854,8 +2854,8 @@ export class Eagle {
             }
 
             // validate edge
-            const isValid: Eagle.LinkValid = Edge.isValid(this, edge.getId(), edge.getSrcNodeKey(), edge.getSrcPortId(), edge.getDestNodeKey(), edge.getDestPortId(), edge.isLoopAware(), edge.isClosesLoop(), false, true, null);
-            if (isValid === Eagle.LinkValid.Impossible || isValid === Eagle.LinkValid.Invalid || isValid === Eagle.LinkValid.Unknown){
+            const isValid: Edge.Validity = Edge.isValid(this, edge.getId(), edge.getSrcNodeKey(), edge.getSrcPortId(), edge.getDestNodeKey(), edge.getDestPortId(), edge.isLoopAware(), edge.isClosesLoop(), false, true, null);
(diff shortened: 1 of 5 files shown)


M-Wicenec (changes requested):
throwing away a lot of code, great!
performance seems fine here too
one change that may be needed is to allow edges with warnings, as it is it seems very strict on what it will deem a valid auto-suggest

M-Wicenec:
Andreas agreed, warnings should be allowed through the auto suggest

james-strauss-uwa:
This should be ready for review again.

I've added a setting for the "Auto-complete edges level" in the "Advanced Editing" tab, which lets users specify whether they want only valid, warnings, or invalid edges, to be suggested by the auto-complete system.

This setting defaults to only show valid edges for Student and Minimal mode. Show Warning edges too for the other modes, except for Expert mode, which shows Invalid edges too. I have no problem changing these defaults if you have feedback.

One other change was required:

- I wanted the defaults for the new setting to use the existing values in the Eagle.LinkValid enum (Valid, Invalid, Warning etc)
- But importing Eagle from Settings.ts caused a import dependency loop, since Eagle imports Settings first.
- So, I did what I had been thinking of for a while, and moved the Eagle.LinkValid enum to Edge.Validity. No reason for that enum to be in Eagle, it just concerns Edges.
- Now it can be imported more easily, in this code and it at least one other spot.

M-Wicenec (approved):
works great. ship it!

PR #716, opened 2024-07-01
Better delete message that doesn't 'double count' children of selected objects

james-strauss-uwa:
## Summary by Sourcery

This pull request addresses a bug in the deletion process by ensuring that child nodes of selected objects are not double-counted. The changes involve checking if child nodes and edges are already in the selected objects or child nodes list before adding them.

- **Bug Fixes**:
    - Fixed issue where child nodes of selected objects were being double-counted in the deletion process.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -3324,7 +3324,29 @@ export class Eagle {
         // find child nodes
         for (const object of data){
             if (object instanceof Node){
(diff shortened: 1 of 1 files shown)


james-strauss-uwa:
Damn, I was clicking buttons on auto-pilot and merged it without thinking.

PR #718, opened 2024-07-03
EAGLE-1259: Error creating gather from selection after undo

james-strauss-uwa:
The selectedObjects array was not updated when reverting to a previous undo snapshot. So the objects in the selectedObjects array were actually from the graph before the undo. When they were updated with a new parent, the changes did not have an effect in the "real" current graph.

## Summary by Sourcery

This pull request addresses a bug where selected objects were not updated correctly after an undo operation, leading to references to non-existent objects. It also includes a refactor in the LogicalGraph class to use a consistent method for accessing nodes.

- **Bug Fixes**:
    - Fixed issue where selected objects were not updated correctly after an undo operation, causing references to non-existent objects.
- **Enhancements**:
    - Refactored LogicalGraph class to use 'nodes()' method consistently instead of 'getNodes()' for better code clarity and consistency.


M-Wicenec (approved):
all good!
would be nice to add that Hierarchy.updateDisplay() function when undo is done, so we keep that display up to date as well.

james-strauss-uwa:
I'll create a ticket for this additional fix.

PR #720, opened 2024-07-04
EAGLE-1194: Preparation and cleanup before actual work

james-strauss-uwa:
This PR doesn't actually do the work of eagle-1194, its just some preparation before starting. I'll create a eagle-1194-2 for the main changes.

I'm planning on starting the work of eagle-1194 by creating a GraphConfig class. This class will store the values of fields that differ from an original graph. 

However, I found that we already had a GraphConfig class. It was used to contain the colours used during graph rendering and a few other const(s) for node radii etc.

I decided to move that into a new file called EagleConfig.ts.

Then I found that we also had Config.ts, and Colours.ts. Which both stored similar things. So I decided to move them into EagleConfig.ts too. I've deleted Config.ts and Colours.ts.

Also added a warning message to EagleConfig.getColor(name), so that if we query for a name that doesn't exist, we get a warning.

Also, there was some code to overwrite CSS variables (using the colours) that was running in the init() function of the nodeRenderHandler. This mean the code ran every time a new node was created. I've moved this into a new function EagleConfig.initCSS() and just run it once on EAGLE startup.

I re-ordered a few import sections, sorting the import alphabetically.

Moved Config.UNDO_MEMORY_SIZE to Undo.MEMORY_SIZE. This is the number of snapshots in the undo history.


M-Wicenec (approved):
nice re-organisation, and good catch on the css initialisation. 
besides the remaining references to graphConfig in the comment_node.html and construct_node.html files, it looks good!

PR #722, opened 2024-07-09
Check and use alternate docker compose option

### LINKED ISSUE #721: `docker-compose` no longer recommended install option
# Issue

`build_eagle.sh` uses the `docker-compose` command for setting up EAGLE on a local machine. It looks like this has been deprecated in favour of `docker compose`: https://docs.docker.com/compose/install/. For newer installs/users of docker/EAGLE, this will break the build script. 

# Potential solution 

It's trivial to fix this locally (just remove the hyphen in favour of a space) so isn't high priority.
If we want to change this in the long run, for backwards compatibility we could check if `docker-compose` exists and change the command accordingly: 

```bash
DOCKER = "docker-compose"

if ! command -v DOCKER
then
    DOCKER = "docker compose"
fi
```

myxie:
This aims to resolve #721 by checking to see if the `docker-compose` utility exists on the user's machine; if not, we try and build EAGLE using the newer plugin approach with `docker compose`.  

(Apologies in advance for naiming the branch incorrectly; the problem with having too many repos open at once!).

## Summary by Sourcery

This pull request enhances the build script to support both 'docker-compose' and 'docker compose' commands, ensuring compatibility with different Docker setups.

- **Enhancements**:
    - Added a check to determine if 'docker-compose' is available and fallback to 'docker compose' if not.

### DIFF
--- a/build_eagle.sh
+++ b/build_eagle.sh
@@ -1,23 +1,32 @@
+COMPOSE="docker-compose"
+
+# Check if the user has a different docker compose setup
+if ! command -v $COMPOSE
+then
+    COMPOSE="docker compose"
+fi
+
+
 case "$1" in
     "dep")
         export VCS_TAG=`git describe --tags --abbrev=0|sed s/v//`
         echo "Building EAGLE version ${VCS_TAG}"
         python updateVersion.py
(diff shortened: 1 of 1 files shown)


james-strauss-uwa (approved):
Looks great, thanks.

PR #723, opened 2024-07-12
Eagle 1260

M-Wicenec:


src/Edge.ts
```diff
@@ -521,15 +538,6 @@ export class Edge {
...
-}
-
-export namespace Edge {
-    export enum Validity {
-        Unknown = "Unknown",        // validity of the edge is unknown
-        Impossible = "Impossible",  // never useful or valid
-        Invalid = "Invalid",        // invalid, but possibly useful for expert users?
-        Warning = "Warning",        // valid, but some issue that the user should be aware of
-        Valid = "Valid"             // fine
+        Eagle.getInstance().logicalGraph().findEdgeById(edgeId)?.issues().push({issue:issue, validity:linkValid})
```
james-strauss-uwa, line 541: Could we avoid this lookup of the edge from its id? By passing the edge (instead of edgeId) into isValidLog()

src/Field.ts
```diff
@@ -943,13 +967,14 @@ export class Field {
...
+                field.issues().push({issue:issue,validity:Errors.Validity.Warning})
+                // errorsWarnings.warnings.push(issue);
             }
         }
 
-        field.errorsWarnings(errorsWarnings)
+        // field.errorsWarnings(errorsWarnings)
 
-        return errorsWarnings
+        // return errorsWarnings
```
james-strauss-uwa, line 977: I think you can remove all the commented code in this function.

All the "errorsWarnings.warnings.push()" and "errorsWarning.errors.push()" etc throughout the function. And these two commented lines at the end.

james-strauss-uwa (approved):
Looks great, just a few minor suggestions.

PR #724, opened 2024-07-16
more helpful port suggestions while dragging

M-Wicenec:
suggested ports, graph-wide will once again be highlighted purple and their name is displayed.
when close enough to a suggested port, the dragging edge will now also be coloured depending on validity

## Summary by Sourcery

This pull request enhances the port dragging experience by providing better visual cues. Suggested ports are now highlighted in purple with their names displayed, and the dragging edge color changes based on the connection's validity.

- **Enhancements**:
    - Enhanced port dragging functionality by highlighting suggested ports in purple and displaying their names.
    - Improved visual feedback during port dragging by coloring the dragging edge based on the validity of the connection.

### DIFF
--- a/src/GraphRenderer.ts
+++ b/src/GraphRenderer.ts
@@ -297,7 +297,8 @@ export class GraphRenderer {
 
     static portDragSuggestedNode : ko.Observable<Node> = ko.observable(null);
     static portDragSuggestedField : ko.Observable<Field> = ko.observable(null);
-    static matchingPortList : {field:Field,node:Node}[] = []
+    static portDragSuggestionValidity : ko.Observable<Errors.Validity> = ko.observable(Errors.Validity.Unknown)
+    static matchingPortList : {field:Field,node:Node,validity: Errors.Validity}[] = []
     static portMatchCloseEnough :ko.Observable<boolean> = ko.observable(false);
 
     //node drag handler globals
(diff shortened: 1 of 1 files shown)


src/GraphRenderer.ts
```diff
@@ -297,7 +297,8 @@ export class GraphRenderer {
 
     static portDragSuggestedNode : ko.Observable<Node> = ko.observable(null);
     static portDragSuggestedField : ko.Observable<Field> = ko.observable(null);
-    static matchingPortList : {field:Field,node:Node}[] = []
+    static portDragSuggestionValidity : ko.Observable<Errors.Validity> = ko.observable(Errors.Validity.Unknown) // this is necessary because we cannot keep the validity on the ege as it does not exist
```
james-strauss-uwa, line 300: Just a minor misspelling in the comment here. "ege" instead of "edge".

src/GraphRenderer.ts
```diff
@@ -297,7 +297,8 @@ export class GraphRenderer {
...
+    static matchingPortList : {field:Field,node:Node,validity: Errors.Validity}[] = []
```
james-strauss-uwa, line 301: Not your fault, since this was the existing variable name. But, like we were discussing last week, I think it is good practice to remove "type" information from the variable names.

Better to call this "matchingPorts" instead of "matchingPortList". No worries if you want to leave it though, its not a change you made.

james-strauss-uwa (approved):
Looks great, nice improvement.

I found a few issues when testing, but nothing resulting from these change. So I'll add the issues to a separate ticket, and Approve this one.

PR #733, opened 2024-08-15
Eagle 1285

M-Wicenec:
The status bar is a new section at the bottom of eagle, allowing us to display relevant actions and shortcuts. This helps with learning eagle and will aid us in removing the inspector from the right window.

New file has been added with an array and system to allow for entries into the "status bar". 
each entry consists of an action/shortcut, a description and a visibility. The visibility will likely be a function to check for a certain scenario like, is a node selected. in which case we could show the action of deleting the node by pressing backspace.

## Summary by Sourcery

Add a new status bar feature to the Eagle interface for displaying actions and shortcuts, and enhance the user interface with transition effects for side windows. Simplify the keyboard shortcut processing logic.

New Features:
- Introduce a status bar at the bottom of the Eagle interface to display relevant actions and shortcuts, enhancing user learning and interaction.

Enhancements:
- Optimize the `getKeyboardShortcutTextByKey` function to return only the first shortcut option for a given key, simplifying the logic and output.
- Add transition effects to the status bar when toggling the visibility of side windows, ensuring smooth movement.


src/StatusEntry.ts
```diff
@@ -0,0 +1,50 @@
+import { Eagle } from './Eagle';
+import * as ko from "knockout";
+import { Setting } from './Setting';
+import { Utils } from './Utils';
+
+export class StatusEntry {
+    action:string;
+    message:string;
+    visibility:ko.Observable<boolean>;
```
james-strauss-uwa, line 9: This works without using an observable for "visibility".

james-strauss-uwa (approved):
Looks good, just those two things we spoke about.

PR #735, opened 2024-08-20
Eagle 1269 - Remove keys from EAGLE

james-strauss-uwa:
This work removes keys from EAGLE, and uses ids (uuidv4) everywhere instead. It touches many systems, but simplifies the code a lot, particularly when loading data from JSON.

Edge now contains: id, srcNodeId, destNodeId instead of _id, srcNodeKey, destNodeKey
Field now contains: nodeId instead of nodeKey
Node now contains: id, parentId, embedId instead of _id, key, parentKey, embedKey

The ids have been given more specific types (defined in main.ts). Instead of all ids being of type "string", they are now "NodeId", "FieldId" or "EdgeId". New ids can be generated using Utils.generateNodeId(), Utils.generateFieldId() or Utils.generateEdgeId(). This prevents using the wrong type of id in a function.

This change modifies the JSON format significantly, so a new function, updateKeysToIds(), is added to GraphUpdater, and called whenever EAGLE loads JSON.

It would be useful if you could load any test graphs that you're familiar with, to check that they load correctly.

## Summary by Sourcery

Transition from using keys to UUIDs for nodes, fields, and edges across the system to enhance data integrity and type safety. Update JSON format and utility functions to support the new ID system, and adjust tests and documentation accordingly.

Enhancements:
- Replace keys with UUIDs for nodes, fields, and edges to improve data consistency and prevent type mismatches.
- Introduce specific types for IDs such as NodeId, FieldId, and EdgeId to enhance type safety.
- Add utility functions to generate new IDs, ensuring unique identifiers across the system.
- Update JSON format to use IDs instead of keys, simplifying data handling and improving code readability.

Documentation:
- Update documentation to reflect changes from keys to IDs, ensuring users understand the new data structure.

Tests:
- Modify tests to accommodate the transition from keys to IDs, ensuring all test cases are compatible with the new ID system.


M-Wicenec (approved):
looks good, did quite a bit of testing, havent found any more bugs. we do need to think about shortening the ids in the future, and work on systems that remove the need of having to show the ids in the first place. Great work, lots of changes!

PR #739, opened 2024-09-18
Eagle 1301

M-Wicenec:
storing right window mode, right and left window visibility as a hidden setting. this means these ui items will continue as you left off when you reload the page.

each of these ui elements are now also using and referencing the setting directly, this means i was able to remove some code from the SideWindow class.

## Summary by Sourcery

Persist window mode and visibility settings for the right and left windows, allowing them to maintain their state across page reloads. Refactor the SideWindow class to utilize settings for managing window visibility, removing redundant code.

Enhancements:
- Store the right window mode and the visibility of both right and left windows as hidden settings, allowing UI elements to persist their state across page reloads.
- Refactor the SideWindow class to remove the shown property and instead use settings to manage window visibility, simplifying the codebase.


src/SideWindow.ts
```diff
@@ -3,26 +3,47 @@ import * as ko from "knockout";
 import { Eagle } from './Eagle';
 import { Utils } from './Utils';
 import { Setting } from "./Setting";
+import { UiModeSystem } from "./UiModes";
 
 export class SideWindow {
-    shown : ko.Observable<boolean>;
```
james-strauss-uwa, line 8: At first I wasn't sure why you had removed shown but kept width. Even though both have been moved to settings.

Then I remembered our discussion about how width changes at high frequency, for example when dragging the window.

Maybe just put a quick comment next to width, explaining why it remains, why it requires a different pattern than 'shown'.

james-strauss-uwa (approved):
Looks good. Thanks.

PR #743, opened 2024-09-30
EAGLE-1309: Improvements to KeyboardShortcut

james-strauss-uwa:
- Added per-key modifier, so a single shortcut can use, for example Ctrl-X and Alt-Y
- Renamed the "key" attribute (e.g. toggle_left_window) to be the "id" attribute, so disambiguate from the actual keyboard key that triggers the shortcut
- Added "warnWhenCantRun" boolean. If true, notifies the user when the "canRun" function returns false
- Added "inputOK" boolean. If true, the shortcut can run when an input or textarea element is focussed
- Renamed "quickAction" to "quickActionDisplay", and "display" to "shortcutListDisplay", to cover the two different functions determining whether a shortcut should be displayed.
- Factored out a single function where the text for a shortcut (e.g. "Ctrl-X") is generated
- Added a couple of "helper" functions, QUICK_ACTION and QUICK_ACTION_DOCS that make specifying QuickActions more compact.
- Added code to skip evaluation and display of shortcuts that don't "work" on a particular platform (e.g. "Alt" on a mac, and "Ctrl" on windows).
- Moved Utils.getKeyboardShortcutTextByKey() to KeyboardShortcut.idToText()
- Some keyboard shortcuts would call functions like "edgeIsSelected" in their canRun function. I've removed these calls in favour of allowing the shortcut to run, and then just notify the user that no edge is selected within the code called by the shortcut.

## Summary by Sourcery

Enhance the KeyboardShortcut functionality by adding per-key modifiers, new boolean properties for user notifications and input focus handling, and refactoring the shortcut text generation. Rename attributes for clarity and improve platform-specific shortcut handling. Introduce helper functions for QuickActions and streamline the shortcut display logic.

New Features:
- Introduce per-key modifiers to allow shortcuts to use different keys with modifiers like Ctrl and Alt.
- Add a 'warnWhenCantRun' boolean to notify users when a shortcut cannot be executed.
- Add an 'inputOK' boolean to allow shortcuts to run when an input or textarea element is focused.
- Implement helper functions QUICK_ACTION and QUICK_ACTION_DOCS to simplify the specification of QuickActions.

Bug Fixes:
- Remove calls to functions like 'edgeIsSelected' in canRun functions of shortcuts, allowing them to run and notify users if conditions are not met.

Enhancements:
- Rename the 'key' attribute to 'id' to avoid confusion with the actual keyboard key triggering the shortcut.
- Rename 'quickAction' to 'quickActionDisplay' and 'display' to 'shortcutListDisplay' for clarity in shortcut display logic.
- Refactor the generation of shortcut text into a single function for consistency.
- Skip evaluation and display of shortcuts that are not applicable on certain platforms.
- Move Utils.getKeyboardShortcutTextByKey() to KeyboardShortcut.idToText() for better organization.


M-Wicenec (approved):
great revamp, OS differentiation works too!

PR #746, opened 2024-10-03
EAGLE-1276: Use File instead of Memory when connecting an Application to a BashShellApp

james-strauss-uwa:
When connecting two applications, EAGLE automatically adds a Memory node in the middle. This is OK most of the time, except when the destination node is a BashShellApp. In that case, the edge would be an error, since the BashShellApp can't read data from a memory location.

Instead, we should use a File node.

Also, adds a (missing?) check to the Edge isValid function, that generates an error if an Application is connected directly to an Application.

## Summary by Sourcery

Modify the connection logic to use a File node instead of a Memory node when connecting to a BashShellApp to prevent errors due to BashShellApp's inability to read from memory.

Bug Fixes:
- Fix the connection issue by using a File node instead of a Memory node when the destination application is a BashShellApp, as BashShellApp cannot read from memory.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -4649,15 +4649,24 @@ export class Eagle {
             y: (srcNodePosition.y + (numIncidentEdges * PORT_HEIGHT) + destNodePosition.y + (numIncidentEdges * PORT_HEIGHT)) / 2.0
         };
 
-        const memoryComponent = Utils.getPaletteComponentByName(Category.Memory);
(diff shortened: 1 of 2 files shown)


M-Wicenec (approved):
works great. 
ive created a separate ticket i will look at about Eagle complaining about the app to app connection with a pop up notification but then fixing it. this notification shouldnt be shown as an error

PR #752, opened 2024-11-01
Eagle 1320

M-Wicenec:
graph configs are much more open to user changes now
i have removed the restrictions stopping users from editing existing configs
the active graph config is now not a separate entity, we keep an id for the active graph config. editing and adding of configs is all done directly into the logical graph, meaning there is also no longer a need to save graph configs before saving the graph

## Summary by Sourcery

Enhance the graph configuration system by allowing direct editing and addition of configurations into the logical graph, eliminating the need for separate saving. Track the active graph configuration using an ID instead of a separate entity.

Enhancements:
- Allow users to edit and add graph configurations directly into the logical graph without needing to save them separately.
- Remove the need for a separate active graph configuration entity by using an ID to track the active configuration.


james-strauss-uwa (approved):
Nice work. This makes graph configs SO much easier to deal with.

I just have a couple of comments and a couple of suggestions.

PR #756, opened 2024-11-04
EAGLE-1337: Fixed bug when loading graphs with no graph configs

james-strauss-uwa:
Fixed bug where code would try to set last graph config as active, even when no graph configs exist.

Now the code checks that at least one graph config exists. If none exist, then graph.activeGraphConfigId is set to undefined.

## Summary by Sourcery

Bug Fixes:
- Fix bug where the application attempts to set the last graph configuration as active even when no graph configurations exist by checking if at least one configuration is present before setting the active configuration ID.


M-Wicenec (approved):
nice catch

PR #758, opened 2024-11-05
EAGLE-1335: New steps in checkGraph to check correctness of the graphConfigs

james-strauss-uwa:
I've added three things:

- Added graphConfigs to the list of things already checked (node, edge, field etc) to make sure all ids are unique
- Make sure the activeGraphConfigId is actually an id that belongs to a graphConfig in the graph
- Make sure that all fields, in all nodes, in all graphConfigs can actually be found in the LogicalGraph

Fixed bug that happened when you added a field to a graph config, then deleted the field from the LG node, or deleted the whole LG node. The graph config would then refer to a field/node that didn't exist, which would break the GraphConfigurationTable.

Now EAGLE shows a placeholder row for the "missing" field. See screenshot:

Screenshot 2024-11-05 at 3 23 49 PM

Also fixed a bug where the part of checkGraph that checked the LG itself was not called as part of Utils.checkGraph()

!!! **Please check** the way I modified ParameterTable.getNodeLockedState() to make the names in the GraphConfigurationAttributesTable readonly. Not sure if this is the best way.

## Summary by Sourcery

Enhance the checkGraph function to include validation for graphConfig IDs and ensure activeGraphConfigId is valid. Fix bugs related to non-existent field and node references in graph configurations, and improve the ParameterTable to display placeholders for missing fields.

New Features:
- Introduce checks to ensure all graphConfig IDs are unique and valid within the LogicalGraph.

Bug Fixes:
- Fix a bug where graph configurations could reference non-existent fields or nodes, causing errors in the GraphConfigurationTable.
- Resolve an issue where the checkGraph function did not validate the LogicalGraph itself.

Enhancements:
- Add placeholder rows for missing fields in the ParameterTable to handle cases where fields are deleted from the LogicalGraph.

### DIFF
--- a/src/LogicalGraph.ts
+++ b/src/LogicalGraph.ts
@@ -901,13 +901,11 @@ export class LogicalGraph {
         const eagle = Eagle.getInstance()
         const graph = eagle.logicalGraph()
 
-        // check that all node, edge, field ids are unique
-        // {
+        // check that all node, edge, field, and config ids are unique
         const ids : string[] = [];
 
         // loop over graph nodes
         for (const node of graph.getNodes()){
-            //check for unique ids
             if (ids.includes(node.getId())){
                 const issue: Errors.Issue = Errors.ShowFix(
                     "Node (" + node.getName() + ") does not have a unique id",
@@ -916,10 +914,10 @@ export class LogicalGraph {
                     "Assign node a new id"
                 );
                 graph.issues.push({issue : issue, validity : Errors.Validity.Error})
-                // errorsWarnings.errors.push(issue);
             }
             ids.push(node.getId());
 
+            // loop over fields within graphs
             for (const field of node.getFields()){
                 if (ids.includes(field.getId())){
                     const issue: Errors.Issue = Errors.ShowFix(
@@ -929,7 +927,6 @@ export class LogicalGraph {
                         "Assign field a new id"
                     );
                     graph.issues.push({issue : issue, validity : Errors.Validity.Error})
-                    // errorsWarnings.errors.push(issue);
                 }
                 ids.push(field.getId());
             }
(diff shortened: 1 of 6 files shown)


src/LogicalGraph.ts
```diff
@@ -901,13 +901,14 @@ export class LogicalGraph {
         const eagle = Eagle.getInstance()
         const graph = eagle.logicalGraph()
 
-        // check that all node, edge, field ids are unique
-        // {
+        // clear old issues
+        graph.issues([]);
+
+        // check that all node, edge, field, and config ids are unique
```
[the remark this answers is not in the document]
james-strauss-uwa: I tried this, but it saved about 3 lines of code overall. Not worth the added complexity.

src/ParameterTable.ts
```diff
@@ -199,6 +204,10 @@ export class ParameterTable {
     // TODO: move to Eagle.ts?
     //       doesn't seem to depend on any ParameterTable state, only Eagle state
     static getNodeLockedState = (field:Field) : boolean => {
+        if (Setting.find(Setting.BOTTOM_WINDOW_MODE).value() === Eagle.BottomWindowMode.GraphConfigAttributesTable){
```
M-Wicenec, line 207: i think this is fine. maybe add a comment that this is for the graph config table, the fields of which are always editable. 
We might have to revisit whether that is the case when we revamp ui-modes/permissions in the future

src/LogicalGraph.ts
```diff
@@ -945,10 +945,78 @@ export class LogicalGraph {
...
+                            function(){
+                                graphConfigNode.removeFieldById(graphConfigField.getId());
+                            },
+                            "Delete field from node in graph config"
+                        );
+                        graph.issues.push({issue: issue, validity: Errors.Validity.Error});
+                    }
+                }
+            }
+        }
```
M-Wicenec, line 1020: i like the way this scenario is handled. The dev console is quite noisy about not being able to find a node though. perhaps we dont need to show those warnings?

M-Wicenec: oh, i also just noticed that when the table is showing the missing node and field in the graph configuration attributes table, the tooltips cause an error when hovering on the node name.

M-Wicenec (changes requested):
nice work! just three minor comments.

src/ParameterTable.ts
james-strauss-uwa: I've added a comment to the code:

`// this handles a special case where EAGLE is displaying the "Graph Configuration Attributes Table"
  // all the fields shown in that table should be locked (readonly)`

This code is meant to make the "Attribute Name" field in the Graph Configuration Attributes Table" readonly. Since it doesn't really make sense to be able to rename a field when looking at a graph config.

src/LogicalGraph.ts
james-strauss-uwa: Yeah, true, there are a lot of warnings.

I'm using findNodeById**Quiet**() here, but there are a few other places that generated warnings. I've now changed the html data-bindings to use findNodeByIdQuiet() too.

And also removed two console.warn() that were explicitly called when getTableFields() could not find a node or field.

M-Wicenec:
I just noticed as well. two of the columns in the graph config attr table disappear when you delete the node and its showing the missing node field. they should probably still be visible so you can tell by the comment what it was, and are able to delete it yourself from there?

image

src/LogicalGraph.ts
james-strauss-uwa: The tooltip error was caused by passing the eagleTooltip binding undefined html.

I've updated the eagleTooltip code to avoid that error.

james-strauss-uwa:
> I just noticed as well. two of the columns in the graph config attr table disappear when you delete the node and its showing the missing node field. they should probably still be visible so you can tell by the comment what it was, and are able to delete it yourself from there?
> 
> image

Thanks, I've just fixed this error. The two columns are now visible and contain correct information where possible.

I've also fixed the "remove" button in the "remove" column, so that it won't error when removing the field, and will re-check the graph after removing the field.

james-strauss-uwa:
I think I've addressed all the issues you found, so it should be ready for another review.

M-Wicenec (approved):
yep, looks all good now

PR #761, opened 2024-11-07
EAGLE-1331: Fix bug where QuickActions with no keyboard shortcut would still display "[]"

james-strauss-uwa:
Removed 'empty' keys from QuickAction helper functions.
Updated KeyboardShortcut.getText() to return empty string if no shortcut.

## Summary by Sourcery

Improve the KeyboardShortcut class by removing unnecessary 'empty' keys from QuickAction helper functions and updating getText() to return an empty string when no shortcuts are defined.

Bug Fixes:
- Fix KeyboardShortcut.getText() to return an empty string instead of '[]' when no key texts are produced.

Enhancements:
- Remove 'empty' keys from QuickAction helper functions to streamline the creation of keyboard shortcuts.

### DIFF
--- a/src/KeyboardShortcut.ts
+++ b/src/KeyboardShortcut.ts
@@ -70,6 +70,12 @@ export class KeyboardShortcut {
             texts.push(result);
         }
 
+        // if there are no key texts produced, then it doesn't make sense to return "[]",
+        // just return an empty string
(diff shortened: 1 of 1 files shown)


M-Wicenec (approved):
nice catch

PR #766, opened 2024-11-08
EAGLE-1344: Developer setting to apply active graph config to the graph before translation

james-strauss-uwa:
Just a temp option until DALiuGE supports applying configs itself

## Summary by Sourcery

New Features:
- Introduce a developer setting to apply the active graph configuration to the logical graph before translation.


M-Wicenec (approved):
nice, works great!

PR #770, opened 2024-11-11
fixed the issue with the status bar and small screens

M-Wicenec:
if the width of the status bar is too small to fit the content, now it will have a horizontal scroll bar.
You can test this by changing the size of the left or right windows.

## Summary by Sourcery

Bug Fixes:
- Add horizontal scroll bar to the status bar for small screens to ensure content visibility.

### DIFF
--- a/static/base.css
+++ b/static/base.css
@@ -1801,8 +1801,18 @@ ul.nav.navbat-nav .btn-outline-secondary:hover{
     visibility: hidden;
     padding-left: 3px;
     margin-bottom: -0.5px;
-    pointer-events: none;
+    overflow-x: auto;
+    overflow-y: hidden;
 }
+
+#statusBar #scrollContainer{
+    width: max-content;
+}
+
+#statusBar::-webkit-scrollbar{
+    height: 2px;
+}
+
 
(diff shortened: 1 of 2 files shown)


static/base.css
```diff
@@ -1801,9 +1801,19 @@ ul.nav.navbat-nav .btn-outline-secondary:hover{
...
-    pointer-events: none;
+    overflow-x: auto;
+    overflow-y: hidden;
+}
+
+#statusBar #scrollContainer{
+    width: max-content;
 }
 
+#statusBar::-webkit-scrollbar{
```
james-strauss-uwa, line 1812: It seems like webkit-scrollbar doesn't work on Firefox.

https://caniuse.com/?search=webkit-scrollbar

But it doesn't cause an issue, just gets ignored, and I can't find a mozilla equivalent. So I think it's OK.

james-strauss-uwa (approved):

static/base.css
M-Wicenec: huh interesting. looks like the firefox scrollbar for that element is perfect though. so luckily it doesnt need that css statement at all

PR #773, opened 2024-11-13
changed a validity test from error to fixable

M-Wicenec:
when connecting application nodes to other application nodes we were showing an error message, even though we automatically fix this by inserting a data node. ive changed the validity to fixable and edited the message slightly

## Summary by Sourcery

Bug Fixes:
- Change the validity check from an error to a fixable state when connecting application nodes to other application nodes.


james-strauss-uwa (approved):
Thanks, this one had been bugging me for a while, but somehow I never created a ticket. I'm glad to see you've done it.

PR #774, opened 2024-11-15
Eagle 1267

M-Wicenec:
## Summary by Sourcery

Enhance the tooltip functionality by allowing certain text to be excluded from markdown conversion using '|||', and update relevant HTML components to utilize this new format.

Enhancements:
- Modify tooltip binding to exclude certain text from markdown conversion by surrounding it with '|||'.
- Update HTML components to use the new tooltip format for file and repository links.


james-strauss-uwa (approved):
Nice, this fixes the problem.

PR #777, opened 2024-11-15
EAGLE-1352: New GraphConfig, created by "favourite"-ing a field, has null id

james-strauss-uwa:
Added generation of GraphConfig id to the GraphConfig constructor, so it's impossible to forget.
Removed some places in the code where the id was set manually.

## Summary by Sourcery

Generate GraphConfig id in the constructor to prevent null ids and remove manual id assignments.

Bug Fixes:
- Ensure GraphConfig instances have a non-null id by generating the id in the constructor.

Enhancements:
- Remove manual id setting for GraphConfig instances throughout the codebase.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -1490,10 +1490,7 @@ export class Eagle {
      */
 
     newConfig = () : void => {
-        // clone existing active config, assign new id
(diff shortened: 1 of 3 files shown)


M-Wicenec (approved):
good catch

PR #781, opened 2024-11-21
Eagle 1313

M-Wicenec:
Request was to also show the configured value in the configured value in the node parameter table. 
If a field is a config field, the node parameter table will now show both the configured value and the graph's value.

## Summary by Sourcery

Enhancements:
- Display both configured and graph values in the node parameter table for config fields.


james-strauss-uwa (approved):
Nice, looks good

PR #793, opened 2024-12-05
EAGLE-1355: Use the same pattern of ids for Repository and RepositoryFile objects

james-strauss-uwa:
Previously, the Repository and RepositoryFile objects used a large random number as an id. The id was generated using:
`Math.floor(Math.random() * 1000000000000)`
This MR changes the ids to follow the same pattern as the Node, Edge, Field, GraphConfig etc, to use a uuidv4

This also required a related change to Repository.DUMMY. This started causing a javascript file load order issue. Instead of being generated at load-time, I've changed this to be generated at run-time, as Repository.dummy().

## Summary by Sourcery

Update the ID generation for Repository and RepositoryFile objects to use uuidv4 for consistency with other components. Adjust the Repository.DUMMY to be generated at runtime to fix a load order issue.

Enhancements:
- Change the ID generation for Repository and RepositoryFile objects to use uuidv4 instead of random numbers.
- Modify the Repository.DUMMY to be generated at runtime as Repository.dummy() to resolve a JavaScript file load order issue.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -912,7 +912,7 @@ export class Eagle {
         const schemaVersion: Daliuge.SchemaVersion = Utils.determineSchemaVersion(dataObject);
 
         const errorsWarnings: Errors.ErrorsWarnings = {errors: [], warnings: []};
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
nice change, looks good!

PR #794, opened 2024-12-05
pull apart parameter table and config parameter table

M-Wicenec:
## Summary by Sourcery

Enhancements:
- Refactor the parameter table to separate the GraphConfigAttributesTable and ParameterTable modes into distinct templates.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -3896,7 +3896,7 @@ export class Eagle {
             $(clickTarget).trigger("select")
 
             //scroll to new row
-            $("#parameterTable .modal-body").animate({
+            $(".parameterTable .modal-body").animate({
                 scrollTop: (fieldIndex*30)
             }, 1000);
         }, 100);
@@ -4169,11 +4169,6 @@ export class Eagle {
             allFieldNames.push(field.getDisplayText() + " (" + field.getType() + ")");
         }
 
-        // if we are summoning this editField modal from the params table, close the params table
-        if (modalType === Eagle.ModalType.Field){
-            $('#parameterTable').modal("hide");
-        }
-
         //if creating a new field
         if (modalType === Eagle.ModalType.Add) {
             $("#editFieldModalTitle").html(this.selectedNode().getName() + " - " + Field.getHtmlTitleText(parameterType, usage));
@@ -4244,7 +4239,7 @@ export class Eagle {
 
                 // if we summoned this editField modal from the params table, now that we are done, re-open the params table
                 if (modalType === Eagle.ModalType.Field){
-                    $('#parameterTable').modal("show");
+                    $('.parameterTable').modal("show");
                 }
             });
         }
@@ -4273,7 +4268,7 @@ export class Eagle {
             clickTarget.focus() //used to focus the field allowing the user to immediately start typing 
             $(clickTarget).trigger("select")
 
-            $("#parameterTable .modal-body").animate({
+            $(".parameterTable .modal-body").animate({
                 scrollTop: (fieldIndex*30)
             }, 1000);
         }, 100);
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -4256,14 +4256,14 @@ export class Eagle {
             fieldIndex = ParameterTable.selectionParentIndex() + 1
             this.selectedNode().addFieldByIndex(copiedField,fieldIndex)
         }else{
-            //if no call in the table is selected, in this case the new node is 
+            //if no cell in the table is selected, in this case the new node is appended at the bottom
             this.selectedNode().addField(copiedField)
             fieldIndex = this.selectedNode().getFields().length -1
         }
 
         setTimeout(function() {
-            //handling selecting and highlighting the newly created node
-            const clickTarget = $($(".paramsTableWrapper tbody").children()[fieldIndex]).find('.selectionTargets')[0]
+            //handling selecting and highlighting the newly created field on the node
+            const clickTarget = $(".paramsTableWrapper tr:nth-child(" + (fieldIndex+1) + ") .selectionTargets")[0]
             clickTarget.click() //simply clicking the element is best as it also lets knockout handle all of the selection and observable update process
             clickTarget.focus() //used to focus the field allowing the user to immediately start typing 
             $(clickTarget).trigger("select")
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -3889,7 +3889,7 @@ export class Eagle {
         //a timeout was necessary to wait for the element to be added before counting how many there are
         setTimeout(function() {
             //handling selecting and highlighting the newly created row
-            const clickTarget = $($("#paramsTableWrapper tbody").children()[fieldIndex]).find('.selectionTargets')[0]
+            const clickTarget = $($(".paramsTableWrapper tbody").children()[fieldIndex]).find('.selectionTargets')[0]
 
             clickTarget.click() //simply clicking the element is best as it also lets knockout handle all of the selection and observable update processes
             clickTarget.focus() // used to focus the field allowing the user to immediately start typing
@@ -4263,7 +4263,7 @@ export class Eagle {
 
         setTimeout(function() {
             //handling selecting and highlighting the newly created node
-            const clickTarget = $($("#paramsTableWrapper tbody").children()[fieldIndex]).find('.selectionTargets')[0]
+            const clickTarget = $($(".paramsTableWrapper tbody").children()[fieldIndex]).find('.selectionTargets')[0]
             clickTarget.click() //simply clicking the element is best as it also lets knockout handle all of the selection and observable update process
             clickTarget.focus() //used to focus the field allowing the user to immediately start typing 
             $(clickTarget).trigger("select")
--- a/static/tables.css
+++ b/static/tables.css
@@ -10,31 +10,28 @@
     background: white;
 }
 
-#tableInspectorSelection{
+#tableInspectorValue, #configTableInspectorValue{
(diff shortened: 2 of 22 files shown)


static/tables.css
```diff
@@ -10,20 +10,22 @@
     background: white;
 }
 
-#tableInspectorSelection{
-    width: 299px !important;
+#tableInspectorValue, #configTableInspectorValue{
```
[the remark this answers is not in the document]
M-Wicenec: its good to have an id on the wrapper for each table for css selectors or functions that should be specific to one table with children that have common classes between the two

M-Wicenec:
i think i got all of it, this was like splitting hairs...

templates/base.html
```diff
@@ -205,13 +205,16 @@
                         </div>
                     </div>
                     <div class="content">
-                        <!-- ko if: Setting.findValue(Setting.BOTTOM_WINDOW_MODE) === Eagle.BottomWindowMode.ParameterTable || Setting.findValue(Setting.BOTTOM_WINDOW_MODE) === Eagle.BottomWindowMode.GraphConfigAttributesTable -->
+                        <!-- ko if: $root.getEagleIsReady() && Setting.findValue(Setting.BOTTOM_WINDOW_MODE) === Eagle.BottomWindowMode.ParameterTable -->
```
james-strauss-uwa, line 208: Maybe you could remove the four $root.getEagleIsReady() checks, and instead use a single check for a "ko if" block around the whole thing.

templates/config_parameter_table.html
```diff
@@ -0,0 +1,179 @@
+<!-- Parameter/arguments table -->
```
james-strauss-uwa, line 1: I think you MIGHT be able to remove all:

`<!-- ko if: Setting.findValue(Setting.BOTTOM_WINDOW_MODE) === Eagle.BottomWindowMode.GraphConfigAttributesTable -->`

within this file. I guess that this whole section of html will only be shown if the BOTTOM_WINDOW_MODE check on line 211 of base.html is 'true'.

So maybe BOTTOM_WINDOW_MODE === GraphConfigAttributesTable is guaranteed to be true within this file?

templates/config_parameter_table.html
```diff
@@ -0,0 +1,179 @@
...
+<div class="eagleTableDisplay" id="parameterTable" tabindex="-1" role="dialog" aria-labelledby="parameterTableTitle">
```
james-strauss-uwa, line 2: Some of these ids (for example: "parameterTableTitle") are re-used in both parameter_table.html and the new config_parameter_table.html.

It is probably best to make the ids unique, even though in practice only one of the elements should be used at any one time.

templates/modals.html
```diff
@@ -46,7 +46,7 @@ <h5 class="modal-title" id="inputModalTitle">Title</h5>
 </div>
 
 <!-- Request User Text Input Modal -->
-<div class="modal fade" id="inputTextModal" tabindex="-1" role="dialog" aria-labelledby="inputTextModalTitle" aria-hidden="true">
```
james-strauss-uwa, line 49: The `aria-hidden="true"` parameter seems to be used in the example modals in the bootstrap docs:

https://getbootstrap.com/docs/5.3/components/modal/#live-demo

Having said that, I don't know what they do? What error did they cause than meant they had to be removed?

templates/parameter_table.html
```diff
@@ -1,11 +1,6 @@
 <!-- Parameter/arguments table -->
 <div class="eagleTableDisplay" id="parameterTable" tabindex="-1" role="dialog" aria-labelledby="parameterTableTitle">
     <div class="bottomWindowHeaderWrap" data-bind="click: function(){ParameterTable.resetSelection()}">
-        <!-- ko if: Setting.findValue(Setting.BOTTOM_WINDOW_MODE) === Eagle.BottomWindowMode.GraphConfigAttributesTable -->
-            <h5 class="bottomWindowHeader">
-                Graph Attributes Configuration Table: <span data-bind="html: $root.logicalGraph().getActiveGraphConfig() ? $root.logicalGraph().getActiveGraphConfig()?.name() : ''"></span>
-            </h5>
-        <!-- /ko -->
         <!-- ko if: Setting.findValue(Setting.BOTTOM_WINDOW_MODE) === Eagle.BottomWindowMode.ParameterTable && eagle.selectedNode() != null -->
```
james-strauss-uwa, line 4: Again here, I think maybe you can assume that `BOTTOM_WINDOW_MODE === Eagle.BottomWindowMode.ParameterTable` within this file.

james-strauss-uwa (changes requested):
I'm glad you're working on this!

I just found a few more things that could be improved.

templates/modals.html
M-Wicenec: yea i saw they were being used, but in practise for eagle, removing them seemed to  change nothing. the error popped up seemingly randomly. it was just a console error, no issues but i figured since they didnt seem to do anything ill just remove them.

templates/config_parameter_table.html
M-Wicenec: good point, yep ill get rid of all of those

templates/modals.html
james-strauss-uwa: I looked it up, it removes the element from the accessibility tree. So only useful for screen readers etc. I'm not sure why bootstrap hides modals from accessibility, surely they would need to be read by a screen reader.

Anyway, it should no be a problem to remove the aria-hidden params.

M-Wicenec:
i think its ready for another look. went a bit nuts removing some more unused stuff and converted the multi used ids into classes

src/Eagle.ts
```diff
@@ -4268,12 +4263,12 @@ export class Eagle {
 
         setTimeout(function() {
             //handling selecting and highlighting the newly created node
-            const clickTarget = $($("#paramsTableWrapper tbody").children()[fieldIndex]).find('.selectionTargets')[0]
+            const clickTarget = $($(".paramsTableWrapper tbody").children()[fieldIndex]).find('.selectionTargets')[0]
```
james-strauss-uwa, line 4266: This comment actually refers to Eagle.ts line 4259.

There are a few errors in the comment: "cell" is misspelled, and the end of the comment seems to be missing.

james-strauss-uwa, line 4266: Also, it may be possible to do this with one CSS selector. Something like:

`const clickTarget = $(".paramsTableWrapper tbody:nth-child(" + fieldIndex + ") .selectionTargets")[0]`

templates/parameter_table.html
```diff
@@ -1,12 +1,7 @@
 <!-- Parameter/arguments table -->
-<div class="eagleTableDisplay" id="parameterTable" tabindex="-1" role="dialog" aria-labelledby="parameterTableTitle">
+<div class="eagleTableDisplay parameterTable" tabindex="-1" role="dialog" aria-labelledby="parameterTableTitle">
```
james-strauss-uwa, line 2: I think we could probably rename this file to be "node_parameter_table.html".

Then we'd have "node parameter table" and "config parameter table" as specific instances of the generic "parameter table".

We'd not only rename the file, but a lot of the classes within.

But renaming everything is a lot of work, so maybe just create another ticket to talk about it/do it later.

james-strauss-uwa (approved):
Just a few minor things, and a few suggestions for larger things that could be done later as separate tickets.

I think once this is merged, we can probably do the release.

src/Eagle.ts
james-strauss-uwa: If it is not possible, or difficult, then don't worry about it. It is just the double $() that spooked me.

templates/parameter_table.html
```diff
@@ -21,28 +16,25 @@ <h5 class="bottomWindowHeader">Fields Table: <span data-bind="html: selectedNode
...
+            <!-- /ko -->
+            <!-- ko if: selectedNode() != null && selectedNode().getCategory() === Category.Docker -->
+                <button class="btn btn-primary float-left" data-bind="click: ParameterTable.initiateBrowseDocker, clickBubble: false, eagleTooltip: `Browse images on dockerhub`" data-bs-toggle="tooltip" data-html="true" data-bs-placement="right" data-bs-original-title="">
+                    Browse DockerHub
+                </button>
             <!-- /ko -->
-            <button type="button" class="btn btn-primary" id="parameterTableAffirmativeButton" data-bind="click:  function(){SideWindow.toggleShown('bottom')}, clickBubble:false"><span id="parameterTableAffirmativeAnswer">Close</span></button>
+            <button type="button" class="btn btn-primary" data-bind="click: function(){SideWindow.toggleShown('bottom')}, clickBubble:false"><span class="parameterTableAffirmativeAnswer">Close</span></button>
```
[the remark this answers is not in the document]
M-Wicenec: ive created a ticket for this

src/Eagle.ts
M-Wicenec: the outer $() in this case is just to turn the return object of whats inside into a jquery object, so we can use .find(). 
ill give your method a try

M-Wicenec: i made it work with this line:
            const clickTarget = $(".paramsTableWrapper tr:nth-child(" + (fieldIndex+1) + ") .selectionTargets")[0]

.children() returns a list of whats inside of the given object. so from your line i had to change "tbody:nth-child" to "tr:nth-child". and i guess .children starts counting at 0 while nth-child starts counting at 1, because i had to add 1 to get to the same item in the list. 
note, fieldIndex is used again after this, so i cant increase the field index before this line

templates/parameter_table.html
M-Wicenec: ive created another ticked for this

PR #798, opened 2024-12-10
EAGLE-1367: Remove orphaned graph config fields when the parent Logical Graph node is removed

james-strauss-uwa:
Remove nodes from Graph Config when the corresponding node in the Logical Graph is removed.

I also improved the display of the "no fields" default message on the graph config. The code no longer checks that a node is selected (this doesn't matter for the graph config). And I added some text to make the status messages a bit more specific.

## Summary by Sourcery

Remove orphaned graph config fields when the corresponding Logical Graph node is deleted and enhance the display of default messages in the graph configuration.

Bug Fixes:
- Remove orphaned graph config fields when the parent Logical Graph node is removed.

Enhancements:
- Improve the display of the 'no fields' default message in the graph config.

### DIFF
--- a/src/GraphConfig.ts
+++ b/src/GraphConfig.ts
@@ -124,6 +124,15 @@ export class GraphConfig {
         }
     }
 
+    removeNodeById = (nodeId: NodeId): void => {
+        for (let i = this.nodes().length - 1; i >= 0 ; i--){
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
nice

PR #799, opened 2024-12-10
EAGLE-1369: Relax warnings on fields with no default value and unknown type

james-strauss-uwa:
Removed check for fields with no default value specified.
Check for known field type now allows anything except undefined or empty string.

## Summary by Sourcery

Relax validation rules by removing the check for fields with no default value and allowing any non-empty, defined field type.

Bug Fixes:
- Fix validation to allow any non-empty, defined field type.

Enhancements:
- Remove the check for fields with no default value, relaxing the validation criteria.

### DIFF
--- a/src/Field.ts
+++ b/src/Field.ts
@@ -846,14 +846,6 @@ export class Field {
                 field.issues().push({issue:issue,validity:Errors.Validity.Error})
         }
 
-        /*
-        // check that the field has a default value
-        if (field.getDefaultValue() === "" && !field.isType(Daliuge.DataType.String) && !field.isType(Daliuge.DataType.Password) && !field.isType(Daliuge.DataType.Object) && !field.isType(Daliuge.DataType.Unknown)) {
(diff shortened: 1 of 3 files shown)


src/Field.ts
```diff
@@ -846,11 +846,13 @@ export class Field {
                 field.issues().push({issue:issue,validity:Errors.Validity.Error})
         }
 
+        /*
         // check that the field has a default value
         if (field.getDefaultValue() === "" && !field.isType(Daliuge.DataType.String) && !field.isType(Daliuge.DataType.Password) && !field.isType(Daliuge.DataType.Object) && !field.isType(Daliuge.DataType.Unknown)) {
             const issue: Errors.Issue = Errors.ShowFix("Node (" + node.getName() + ") has a component parameter (" + field.getDisplayText() + ") whose default value is not specified", function(){Utils.showField(eagle, node.getId(),field)}, function(){Utils.fixFieldDefaultValue(eagle, field)}, "Generate default value for parameter");
                 field.issues().push({issue:issue,validity:Errors.Validity.Warning})
         }
+        */
```
M-Wicenec, line 855: did you want to keep this around for now?

M-Wicenec (approved):
nice, just question comment

src/Field.ts
james-strauss-uwa: No, I should get rid of it. I'll do that.

PR #800, opened 2024-12-10
EAGLE-1363: Fixed bug when editing description of field on Node in palette

james-strauss-uwa:
Fixed an issue where a field description was not editable if the containing node was from a palette.

This used to cause an issue:
1. Select a node in a palette
2. View the parameters table
3. Click the 'edit' icon on the description of a field
4. Observe a javascript console error

The error was caused by the previous code always looking for the field's node within the logicalGraph. If the node was in a palette, then it was not found, and the null value caused issues.

## Summary by Sourcery

Bug Fixes:
- Fix an issue where editing the description of a field on a node from a palette caused a JavaScript console error.


M-Wicenec (approved):
nice catch, i never noticed

PR #803, opened 2024-12-12
EAGLE-1374: New 'developer' class of notifications

james-strauss-uwa:
Added new setting 'Show developer notifications'.

Update Utils.showNotification with new param that specifies whether message is intended for developers.

Removed duplicate console.error/console.warn calls from same spots in code where showNotification() was called. Instead just have showNotification() also call console.error/console.warn within it.

Made the "JSON schema validation failure" notification a developer-only message.

## Summary by Sourcery

Add a new 'developer' class of notifications with a corresponding setting to toggle their visibility. Refactor the showNotification function to include developer-specific messages and consolidate logging calls.

New Features:
- Introduce a new 'developer' class of notifications that can be toggled via a setting.

Enhancements:
- Consolidate console.error and console.warn calls within the showNotification function to streamline error and warning logging.

### DIFF
--- a/src/ComponentUpdater.ts
+++ b/src/ComponentUpdater.ts
@@ -11,7 +11,6 @@ export class ComponentUpdater {
 
         // check if any nodes to update
         if (graph.getNodes().length === 0){
-            // TODO: don't showNotification here! instead add a warning to the errorsWarnings and callback()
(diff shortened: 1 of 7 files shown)


M-Wicenec (approved):
nice! glad that this message is finally hidden from the user

PR #808, opened 2024-12-16
EAGLE-1305: Save collapsed/expanded state of builtin palettes to localStorage

james-strauss-uwa:
Built-in palettes now default to collapsed. Two new settings (saved to localStorage) are used to remember the collapsed/expanded state of the two built-in palettes.

Updated tests with a new step that expands the 'Builtin Components' palette.

Added expanded() observable to Palette class. State of this observable is used in KO data-bind's to add necessary classes to html for collapsing and expanding palette accordions.

Added a click handler to the accordion header, so that when the accordion gets collapsed/expanded, the value of the Palette.expanded() observable is kept in sync.

So now palettes can be collapsed/expanded easily from code, by just setting the Palette.expanded() observable.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -50,8 +50,6 @@ User Options
 
 **Confirm Reload Palettes** - Prompt user to confirm when loading a palette that is already loaded.
 
-**Open Default Palette on Startup** - Open a default palette on startup. The palette contains an example of all known node categories.
-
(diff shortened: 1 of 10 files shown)


M-Wicenec (approved):
awesome, works great. you must have done an amazing job for sourceryAI not to suggest random changes!

PR #824, opened 2025-02-26
Eagle 1395

M-Wicenec:
there are a few fixes for the tutorial here

the left and right arrow key shortcuts for toggling the left and right window have been suppressed while in the tutorial as these are used for navigating through the tutorial steps

i have created a new function to find a node in the palette. this will open all the palettes and scroll the palette list to find the node we are requesting in the tutorial step.

there was an additional issue with the step requesting the selection of the newly added 'helloWorldApp' node in the graph. that should also be working as intended again.

adjusted the random placement function for nodes, it will now place the first node to the left of the center of the canvas

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the 'helloWorldApp' node was not being correctly selected in the graph during the tutorial.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -591,9 +591,11 @@ export class Eagle {
      * It resets all fields in the editor menu.
      */
     resetEditor = () : void => {
-        console.log('resetting the editor')
-        this.selectedObjects([]);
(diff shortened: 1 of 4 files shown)


james-strauss-uwa (approved):
Great, the tutorial can now be completed

PR #828, opened 2025-03-12
EAGLE-1399: Remove knockout components

james-strauss-uwa:
These components palette-component, repository-folder, repository-file, hierarchy-node, etc, were never quite perfect. They relied on code in **src/components.ts** translating them from the real Typescript types into types suitable for the component. If an attribute was added to the real type, but that attribute was not added to the translation code, then the attribute would not be available in the knockout bindings.

To fix this issue, the components have been removed.

Where possible, the same functionality was achieved by just using a server-side (flask/jinja) "include" directive. However, that templating engine could not handle includes with includes inside them.

So to fix those case, we use the knockout "template" data binding. The source code is added to base.html once, inside a <script> element. Then where the code needs to be included, we use a knockout "template" binding.

## Summary by Sourcery

Replaces knockout components with server-side includes and knockout templates to improve maintainability and address issues with attribute handling in knockout bindings. Removes the components.ts file and updates templates to use include directives and knockout templates.

Enhancements:
- Replaces knockout components with server-side includes and knockout templates.
- Improves maintainability by removing the need for translation code between Typescript types and knockout bindings.
- Uses knockout "template" data binding to handle includes with includes inside them.
- Removes the components.ts file.
- Updates templates to use include directives and knockout templates.


M-Wicenec (approved):
looks good, good idea to standardise this a little more

PR #830, opened 2025-03-17
EAGLE-1358: Clean up Eagle.editField

james-strauss-uwa:
There was a lot of quite old code here, left over from when:

- when fields were added using a modal
- when fields were edited from the inspector
- when the ParameterTable and GraphConfigurationTable were modal dialogs

NOTE: The keyboard shortcut used in the ParameterTable where "enter" would move the selection to the next table cell, was not working, since the condition for running was always false. I've changed it to always true. Doesn't seem to cause any problems, but keep an eye out.


james-strauss-uwa:
Updated this so that the editField modal now acts on the field itself, meaning changes take place immediately, instead of only when the 'OK' button is used to dismiss the modal.

Also, always run checkGraph after the modal is dismissed, even when the 'OK' button is not pressed.

M-Wicenec (approved):
works great

PR #837, opened 2025-03-21
Eagle 1408

M-Wicenec:
removed minimum 0 restrictions from all integer value fields

## Summary by Sourcery

Enhancements:
- Removes the minimum value restriction of 0 from all integer value fields.


james-strauss-uwa (approved):
Looks good

PR #844, opened 2025-04-03
EAGLE-1422: Local palettes expanded by default

james-strauss-uwa:
Made the local palette loading use more of the remote palette loading code path. When they are handled similarly, then both palettes are expanded by default.

## Summary by Sourcery

Refactor palette loading to use a common code path for local and remote palette loading, ensuring consistent behavior

Enhancements:
- Simplified palette loading logic by extracting common loading mechanism into a shared method `_reloadPalette()`

Chores:
- Reduced code duplication in palette loading process

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -1241,16 +1241,8 @@ export class Eagle {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


M-Wicenec (approved):
works 👍

PR #850, opened 2025-04-11
EAGLE-1431: Stop EAGLE overwriting modelData when loading palettes

james-strauss-uwa:
Also stop EAGLE re-sorting the nodes in a palette after adding nodes from a graph.

Re-named fields in modelData modal from eagleVersion to generatorVersion, add generatorName.

Fixed bug where modelData.downloadUrl was not displayed.

## Summary by Sourcery

Improve model data handling and display in EAGLE by updating metadata fields, preventing unintended overwrites, and fixing display issues

Bug Fixes:
- Fixed an issue where modelData.downloadUrl was not being displayed correctly
- Stopped EAGLE from overwriting modelData when loading palettes

Enhancements:
- Renamed metadata fields from Eagle-specific terms to more generic generator-related terms
- Improved tooltips for model data modal to provide clearer descriptions

Chores:
- Removed automatic re-sorting of nodes in a palette after adding nodes from a graph

### DIFF
--- a/eagleServer/eagleServer.py
+++ b/eagleServer/eagleServer.py
@@ -944,18 +944,9 @@ def open_url_file():
     if not "modelData" in graph:
         graph["modelData"] = {}
 
-    # add the repository information
-    graph["modelData"]["repo"] = ""
-    graph["modelData"]["repoBranch"] = ""
+    # overwrite some modelData information
     graph["modelData"]["repoService"] = "Url"
-    graph["modelData"]["filePath"] = ""
-
-    # add the GitLab file information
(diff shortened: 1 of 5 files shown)


M-Wicenec (approved):
nice, I assume that means were treating the sort function as a conscious decision of wanting to to alter the real order of the nodes in the palette?

james-strauss-uwa:
> nice, I assume that means were treating the sort function as a conscious decision of wanting to to alter the real order of the nodes in the palette?

Yes, we won't sort by default. But still allow the user to sort, in case they decide that makes organisation easier.

PR #851, opened 2025-04-11
EAGLE-1429: Fields on Construct component should not be ports

james-strauss-uwa:
checkGraph() function already picked up this issue. Construct nodes have a maximum of zero ports. But the issue was only detected when the graph was re-checked, which wasn't happening.

So I added a call to eagle.checkGraph() inside the existing fieldUsageChanged() event handler to re-check graph after change to field usage.

Also update the options data-bind in the UI for the "usage" column in the ParameterTable. If the node is a construct, then only the "NoPort" option appears in the usage drop-down.

## Summary by Sourcery

Modify field usage handling for Construct nodes to prevent port creation and ensure graph consistency

Bug Fixes:
- Prevent fields on Construct nodes from being set as ports by restricting usage options
- Automatically re-check graph integrity when field usage changes

Enhancements:
- Add dynamic usage options computation for nodes based on their category type
- Improve field usage change handling to maintain graph validity

Tests:
- Update page model test selectors to match new class names for field usage dropdown

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -298,6 +298,14 @@ export class Node {
         }
     }, this);
 
+    usageOptions : ko.PureComputed<Daliuge.FieldUsage[]> = ko.pureComputed(() => {
+        // fields on construct nodes cannot be ports
+        if (this.categoryType() === Category.Type.Construct){
+            return [Daliuge.FieldUsage.NoPort];
+        }
+        return Object.values(Daliuge.FieldUsage)
+    }, this);
+
     getInputPorts = () : Field[] => {
         const result: Field[] = []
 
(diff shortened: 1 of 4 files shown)


james-strauss-uwa:
@M-Wicenec rightly pointed out that it doesn't make much sense for the "Use As" dropdown to contain only one option for Construct nodes. Every field on a construct node must have "Use As" set to "No Port".

So, I've now hidden that column when a Construct is selected.

This follows the same pattern as the "Encoding" column, which is hidden unless the selectedNode is an Application.

I've also hidden those two columns in the "columns visibility" drop down, in the cases where the selectedNode doesn't allow that column to be displayed anyway.

This is ready to re-test and review.

M-Wicenec (approved):
yep, looks good!

PR #852, opened 2025-04-11
EAGLE-1428: Docker image build failure

james-strauss-uwa:
Changed loop type and added type annotation

## Summary by Sourcery

Enhancements:
- Replace foreach loops with indexed for loops to improve type safety and explicit type handling in folder traversal methods

### DIFF
--- a/src/Repository.ts
+++ b/src/Repository.ts
@@ -99,7 +99,9 @@ export class Repository {
         for (const pathPart of pathParts){
             let foundPathPart = false;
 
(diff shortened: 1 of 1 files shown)


james-strauss-uwa:
Looks like we'll probably remove the "dev" way of building the docker image. So this change should not be required.

james-strauss-uwa:
Looks like we'll probably remove the "dev" way of building the docker image. So this change should not be required.

PR #854, opened 2025-04-16
EAGLE-1439: Don't overwrite unrecognised node categories when loading 

james-strauss-uwa:
Previously, if EAGLE didn't recognise the category of a node, it would replace the category with "Unknown". This resulted in the loss of some information.

It seems better to just warn the user of the issue, and allow them to solve the problem as they see fit.

Changes:
- Don't overwrite a node category if unrecognised
- Use "icon-none" to display nodes with unrecognised categories
- Add a rule to Node.isValid() to check if the category is unrecognised, if so, warn user
- Allow unrecognised categories to be changed in the inspector UI (and put all known categories in the list of categories it can be changed to)

Testing:
I added a test graph, with an unrecognised category, to the EAGLE-1439 JIRA ticket. The node with the unrecognised category is called "array", it is second from the left. If you select it, you can update the category in the inspector. For example, try change it to a Memory, or File node.

Currently not passing CI, but that seems to be due to the broken "HelloWorldApp" issue. Once that is fixed it should pass.

## Summary by Sourcery

Improve handling of unrecognised node categories in EAGLE by preventing automatic overwriting and providing better user guidance

New Features:
- Enable users to manually update nodes with unrecognised categories

Bug Fixes:
- Prevent automatic overwriting of unrecognised node categories with 'Unknown'

Enhancements:
- Add support for displaying nodes with unrecognised categories
- Implement warning mechanism for unrecognised node categories
- Allow changing unrecognised categories in the inspector UI

### DIFF
--- a/src/CategoryData.ts
+++ b/src/CategoryData.ts
@@ -68,7 +68,7 @@ export class CategoryData {
                 canHaveComponentParameters: false,
                 canHaveApplicationArguments: false,
                 canHaveConstructParameters: false,
-                icon: "error",
+                icon: "icon-none",
                 color: "pink",
                 sortOrder: Number.MAX_SAFE_INTEGER,
             };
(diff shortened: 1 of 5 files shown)


M-Wicenec (approved):
👍

PR #857, opened 2025-04-17
EAGLE-1443: Added button on inspector to display node as Json

james-strauss-uwa:
Also removed some code from displayObjectAsJson() that cleared several attributes from modelData. I'm not sure why it would do that.

The icon is "open in new window" which is not perfect, but I couldn't find any "json"-related icons

## Summary by Sourcery

Add functionality to display a selected node as JSON in the inspector interface

New Features:
- Added a new button in the inspector to display the selected node's JSON representation

Enhancements:
- Simplified the displayObjectAsJson() method by removing unnecessary attribute clearing
- Updated JSON serialization to use a consistent indentation from EagleConfig

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -1404,35 +1404,24 @@ export class Eagle {
     }
 
     displayObjectAsJson = (fileType: Eagle.FileType) : void => {
-        let clone: LogicalGraph | Palette;
+        let jsonString: string;
         
         switch(fileType){
             case Eagle.FileType.Graph:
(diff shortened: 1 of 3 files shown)


M-Wicenec:
works well, id probably use some code related icon like one of the ones highlighted below. i quite like the terminal one 
image

james-strauss-uwa:
I like the "Data Object" one. Those curly braces yell JSON to me. I've changed to that one, thanks for the suggestion.

PR #860, opened 2025-04-23
EAGLE-1449: Remove requirement to save 'local' files before translation

james-strauss-uwa:
If the file is 'local', the translate button will continue with the translation without saving.

## Summary by Sourcery

Modify translation workflow to skip saving for local files

Bug Fixes:
- Remove the requirement to save local files before translation, allowing immediate translation of local files

Enhancements:
- Add a check to determine if the file is local before requiring a save operation


M-Wicenec (approved):
nice and easy!

PR #861, opened 2025-04-23
EAGLE-1441: Update the ParameterTable content when using the undo system

james-strauss-uwa:
Made sure the after both undo and redo actions, the Parameter Table is updated with the latest content

## Summary by Sourcery

Enhancements:
- Ensure the ParameterTable is synchronized with the current selected node after undo and redo operations


M-Wicenec (approved):
works, nice

PR #862, opened 2025-04-23
EAGLE-1445: "Display X as JSON" modal should have readonly textarea

james-strauss-uwa:
Added parameter 'readonly' to Utils.requestUserText(). Set param true when displaying JSON for graph/palette/node.

## Summary by Sourcery

Add readonly option to text input modal for JSON display

Bug Fixes:
- Ensure JSON display modals have a read-only textarea to prevent accidental modifications

Enhancements:
- Modify Utils.requestUserText() to support an optional readonly parameter for the input textarea

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -1415,13 +1415,13 @@ export class Eagle {
                 return;
         }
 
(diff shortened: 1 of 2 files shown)


M-Wicenec (approved):
👍

PR #864, opened 2025-04-23
EAGLE-1451: Fixed bug where palette's "modified" flag was set when the Branch node selected 

james-strauss-uwa:
Removed 'change' event handler function from 'default value' dropdown

This fixes the bug, but now means that nothing happens when the "default value" of a "option"-type field is changed. The graph will not be re-checked etc.

I think the fix is preferable to the bug, but not perfect.

## Summary by Sourcery

Remove change event handler from default value dropdown to prevent unintended palette modification

Bug Fixes:
- Fixed an issue where the palette's 'modified' flag was incorrectly set when selecting a Branch node's default value dropdown

Chores:
- Acknowledged that removing the change event handler means the graph will not be re-checked when default values are changed


M-Wicenec (approved):
good catch!

PR #866, opened 2025-04-24
EAGLE-1437: Change of Application Arguments to shorter versions

james-strauss-uwa:
Switched to shorter fieldType names:

- ApplicationArgument -> Application
- ComponentParameter -> Component
- ConstraintParameter -> Constraint
- ConstructParameter -> Construct

Use originals when loading and saving ONLY, otherwise within EAGLE we use the new versions.

Eventually DALiuGE will be updated with the new names.

Decrease width of "Parameter Type" column in Parameter Table.
Newly added parameters are 'Application'-type instead of 'Component'-type

## Summary by Sourcery

Refactor field type names to use shorter, more concise versions while maintaining compatibility with existing data loading and saving mechanisms

New Features:
- Added methods to convert between legacy and new field type enumerations

Enhancements:
- Introduced conversion methods between old and new field type enums to ensure backward compatibility
- Reduced width of 'Parameter Type' column in parameter table

Chores:
- Replaced verbose field type names with shorter equivalents
- Updated all references to field types across multiple files

### DIFF
--- a/e2e/parameterTableAndKeyboardShortcuts.spec.ts
+++ b/e2e/parameterTableAndKeyboardShortcuts.spec.ts
@@ -50,8 +50,8 @@ test('Parameter Tables and keyboard Shortcuts', async ({ page }) => {
   await expect(page.getByRole('row').last().locator('.typesInput')).toHaveValue('Integer')
 
   //change parameter type then check that the value has been changed accordingly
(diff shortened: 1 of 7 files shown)


M-Wicenec (approved):
looks good!

could even go a little slimmer on the type column now, looks good at 130px to me.

we should probably create a ticket to look at putting a little icon or something to differentiate each of the options in the table as there are three options starting with 'C' with similar word lengths.
image

PR #869, opened 2025-04-28
EAGLE-1461: Configured values don't keep type

james-strauss-uwa:
Use Node field type when writing value of GraphConfig field to JSON

## Summary by Sourcery

Bug Fixes:
- Fix loss of data type information when saving graph configuration fields to JSON, ensuring values are stored according to their defined type (e.g., boolean, number) instead of always as strings.


M-Wicenec (approved):
👍

PR #876, opened 2025-05-02
Eagle 1465

M-Wicenec:
- this ticket contains improvements to the filtering system when dragging an edge from a port into empty space to create another node.
- removed exception for node filtering for ports with the name 'dummy' as this was causing recommendations to connect data  nodes to data nodes. (i believe that was a remnant from before we had the object type)
- simplified the filtering code to reduce code and make it more readable.
- unrelated, but i also improved the positioning of the types dropdown in the parameter table. reducing it's height allows bootstrap to expand upwards instead of downwards when we are running out of space at the bottom of the page. and i added some css overwrites to better position the dropdown in relation to it's header

## Summary by Sourcery

Improve node filtering and suggestion system when dragging ports, and refactor related code for better readability and functionality

New Features:
- Enhanced dropdown positioning in parameter table

Bug Fixes:
- Removed special exception for 'dummy' port filtering
- Fixed node recommendation logic to prevent incorrect data node connections

Enhancements:
- Simplified node filtering logic for port drag and drop
- Improved component matching algorithm for node suggestions

Chores:
- Refactored filtering code to reduce complexity and improve readability

### DIFF
--- a/src/GraphRenderer.ts
+++ b/src/GraphRenderer.ts
@@ -1796,7 +1796,6 @@ export class GraphRenderer {
     }
 
     static showUserNodeSelectionContextMenu(){
-        console.log('showusernodes selection context menu')
         const eagle: Eagle = Eagle.getInstance();
 
         //hiding the suggested node edge while the right click menu shows up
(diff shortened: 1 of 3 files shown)


james-strauss-uwa (approved):
Nice

PR #877, opened 2025-05-02
improved create construct /subgraph from selection functions and remo…

M-Wicenec:
…ved some old code

## Summary by Sourcery

Remove node shrinking logic and replace with a new method to center constructs around their children

Enhancements:
- Replaced node shrinking method with a more centralized approach to positioning group nodes

Chores:
- Removed deprecated shrinkNode method from LogicalGraph class


james-strauss-uwa (approved):
This works well

PR #878, opened 2025-05-06
Eagle 1468

M-Wicenec:
- added create construct and graph from selection to multi selection right click menus
- added create construct from selection to status bar when multiple things are selected
- simplified some of the descriptions in status bar entries
- created a new "edit"  section in the graph dropdown, this houses some of the options that were previously in the graph - new section
- removed the add edge function, since this was only used for tests and is no longer needed


james-strauss-uwa (approved):
Looks good to me

PR #879, opened 2025-05-06
EAGLE-1473: Support for JSON, markdown, python in embedded code editor

james-strauss-uwa:
- Switched from CodeMirror 6 to 5. Easier to integrate, easier to use support languages. The theme is not quite the same, so it appears slightly different.
- Now supports javascript/JSON, markdown, python and plain text
- Display Node as JSON, and Display Graph as JSON now use the source editor, in readonly javascript mode.
- Better mode for plain text, using the "null" highlighting

## Summary by Sourcery

Switch from CodeMirror 6 to CodeMirror 5 for improved language support and easier integration in the embedded code editor

New Features:
- Support for javascript/JSON, markdown, python, and plain text in the embedded code editor

Enhancements:
- Switched to CodeMirror 5 for better language mode support
- Improved code editor configuration


M-Wicenec (approved):
nice, this works well!
we should create a ticket to utilise this for description editing with markdown
ive recently merged the right click on nodes with a func_code field for code editing ticket, you might have to merge master into this one first and see if it still works properly

james-strauss-uwa:
> nice, this works well! we should create a ticket to utilise this for description editing with markdown ive recently merged the right click on nodes with a func_code field for code editing ticket, you might have to merge master into this one first and see if it still works properly

Yes, I'll create a JIRA ticket for description editing.

It seems like github says it can merge automatically, so hopefully that works.

PR #880, opened 2025-05-07
EAGLE-1472: Missing 'num_of_copies' field on Scatter created by 'Create Construct From Selection'

james-strauss-uwa:
Updated both 'Create Construct from Selection' and 'Create Subgraph from Selection' to clone the new parent (construct or subgraph) from that node in the 'Component Template' palette, instead of generating an empty component from scratch.

## Summary by Sourcery

Improve node creation for 'Create Construct from Selection' and 'Create Subgraph from Selection' by cloning existing palette components instead of generating empty nodes

New Features:
- Add fallback to basic node creation if no matching palette component is found

Enhancements:
- Modify node creation to clone existing palette components when available


M-Wicenec (approved):
nice

PR #884, opened 2025-05-09
Eagle 1415

M-Wicenec:
moved the headers of data nodes to the bottom of the node.
just have a quick look at if the distance of the header from the node is behaving correctly

## Summary by Sourcery

Enhancements:
- Adjusted CSS positioning to move data node headers to the bottom of the node

### DIFF
--- a/static/graph.css
+++ b/static/graph.css
@@ -409,7 +409,7 @@
     
 }
 
-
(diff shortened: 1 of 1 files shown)


james-strauss-uwa (approved):
Looks good

PR #885, opened 2025-05-09
EAGLE-1478: Remove graph check for Memory Nodes as input to a BashShellApp

james-strauss-uwa:
Memory DROPs as input to BashShellApps now works in DALiuGE. So, on the EAGLE side we remove the graph check for that condition.

This also has implications when the user drags an edge between two Application nodes. EAGLE used to check whether the destination node was a BashShellApp, and in that case use a File node instead of a Memory.

This change adds a new Setting, in the "Advanced Editing" tab, which allows the user to specify which category of node they would like to use for the intermediate node between two Applications.

The user can choose from all data nodes that can have at least one input and one output.

## Summary by Sourcery

Remove graph checks and restrictions related to Memory Nodes as inputs to BashShellApps, and introduce a new user-configurable setting for selecting intermediate data nodes between applications

New Features:
- Add a new setting in Advanced Editing to specify the default intermediate data node type between Application nodes

Enhancements:
- Dynamically determine intermediate data node type based on user preference
- Allow selection from all data nodes that can have at least one input and one output

Chores:
- Remove hard-coded graph checks for Memory Nodes and BashShellApps
- Update category data to support dynamic intermediate node selection


M-Wicenec (approved):
looks good

PR #886, opened 2025-05-09
more descriptive graph config parameters table state if there is no config

M-Wicenec:
in the config parameter table it will now say you should create a config if there is none (previously it said you should add a field in this case)
additionally ive added a button to quickly create a new config next to this message. This has been added to both the config parameter table and the configs table

## Summary by Sourcery

Improve user experience for graph configurations by adding more descriptive messaging and quick-add buttons when no configurations exist

New Features:
- Added a direct 'Add New Config' button in both config parameter and graph configurations tables when no configurations are present

Enhancements:
- Updated empty state messaging to be more clear about the lack of graph configurations
- Added styling for the new configuration creation button

### DIFF
--- a/static/base.css
+++ b/static/base.css
@@ -543,10 +543,6 @@ color: #00bb00;}
     font-size: 16px;
 }
 
-.parameterTable .tableAddNewConfigBtn, .eagleTableDisplay .tableAddNewConfigBtn{
-    background-color: #0059a5;
-}
-
 #editFieldModal .dropdown {
     width:100%;
 }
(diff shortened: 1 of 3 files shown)


static/base.css
```diff
@@ -543,6 +543,10 @@ color: #00bb00;}
     font-size: 16px;
 }
 
+.parameterTable .tableAddNewConfigBtn, .eagleTableDisplay .tableAddNewConfigBtn{
```
james-strauss-uwa, line 546: Another place in the CSS (.rightWindow .btn-primary) you override both the background-color and the border-color.

Maybe you want to do that here? Or combine these same CSS rules.

james-strauss-uwa (changes requested):
A good change. I think there is just a slight change you could make to the CSS

static/base.css
M-Wicenec: good catch, ive combined the rules

PR #887, opened 2025-05-09
EAGLE-1452: Unit test of graph validation against schema

james-strauss-uwa:
Loads three different graphs, writes them to JSON, and validates the output against the DALiuGE schema.

I chose the three graphs randomly, so we can change them anytime.

## Summary by Sourcery

Add graph validation functionality to EAGLE, allowing users to validate loaded graphs against the DALiuGE JSON schema

New Features:
- Implement graph validation method in Eagle class
- Add 'Validate' option to graph dropdown menu

Enhancements:
- Integrate JSON schema validation with user notification system

Tests:
- Create end-to-end test for graph validation using three different graph examples

### DIFF
--- a/e2e/validatingGraphs.spec.ts
+++ b/e2e/validatingGraphs.spec.ts
@@ -18,17 +18,20 @@ test('Validating Graphs', async ({ page }) => {
     await page.goto('http://localhost:8888/?tutorial=none&service=Url&url='+graphUrl);
 
     // wait for the 'graph load success' notification to be shown
-    await page.waitForSelector('div[data-notify="container"]');
+    await page.locator('div[data-notify="container"]').waitFor({state: 'attached'});
 
     // dismiss notification
     await page.locator('button[data-notify="dismiss"]').click();
+
+    // wait for the 'graph load success' notification to be dismissed
+    await page.locator('div[data-notify="container"]').waitFor({state: 'detached'});
 
     // navigate menus and click the "validate" item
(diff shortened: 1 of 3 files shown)


e2e/validatingGraphs.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  // examples
+  "https://raw.githubusercontent.com/ICRAR/EAGLE-graph-repo/refs/heads/master/examples/HelloWorld-simple.graph",
+
+  // graph_patterns
+  "https://raw.githubusercontent.com/ICRAR/EAGLE-graph-repo/refs/heads/master/graph_patterns/LoopWithBranch.graph"
+]
+
+test('Validating Graphs', async ({ page }) => {
+  page.setViewportSize({width:1920,height:1080})
```
M-Wicenec, line 15: playing around with this branch i noticed we dont seem to need this line. its also used in some of the other tests but this is already set as default so we dont need to re-set it. Additionally running this locally in headed mode(showing browser) i noticed the bottom of eagle is cut off when this line is enabled, after disabling it, it looks normal. maybe should remove these.

M-Wicenec (approved):
nice, just added one comment regarding setViewportSize

e2e/validatingGraphs.spec.ts
james-strauss-uwa: OK, I'll remove it. Thanks.

PR #894, opened 2025-05-14
EAGLE-1427: Graph/Node description markdown editing

james-strauss-uwa:
Added new modal for editing markdown with a side-by-side editor and preview. Modal has two "modes":
- editing
- preview-only

Mode state is remembered in a hidden setting.

Removed markdown support from other requestUserCode() function, since the new requestUserMarkdown() should be used instead.

Other misc fixes:
- commitToGit had the check for "allow graph editing" around the wrong way, preventing commits for users with permission
- replaced "func_code" with Daliuge.FieldName.FUNC_CODE where possible
- removed CSS font-style applied to * that prevented italics from appearing in non-tooltip markdown output


james-strauss-uwa:
@sourcery-ai review

M-Wicenec:
I just noticed you didnt actually request a review on this one, but ive already had a look anyways. so heres some things i noted;

1- when you enter something in the graph description via the new modal, then exit using 'esc', then open the modal again. note the rendered preview is not being shown, until you edit on the left again. This is related to something i find myself doing in some other place in eagle. usually when i edit via a modal such as this, i tend to close it using 'esc' or clicking next to the modal, expecting the changes to apply. yet they only apply when click 'ok'. i think the changes should always apply when you close the modal, its not like there is an option to cancel anyways, since closing it without 'ok' atm leaves the modal in a weird half applied state.

2- when opening the edit code modal, it should reset. related to the point above, this can also be seen when you go into the table on a fresh pyfunc app, edit the default value using the modal and apply with 'ok'. then open the modal on the empty value field. (notice we are shown the value we just entered in the default value field via the modal)

3- now that we have this nice editor, i think it makes sense to show the rendered descriptions in the graph info modal with a little edit button that opens the new edit code modal.

4- minor but would be nice to for the time being at least have an info button in the markup modal, that opens the markup documentation

james-strauss-uwa:
Yes, I requested that sourcery AI re-review this one, but it never did!

I was able to reproduce the bugs in items 1 and 2 in your comment.

I also agree with items 3 and 4, so I'll do those in this branch

james-strauss-uwa:
@M-Wicenec this one should be ready to look at again. I fixed the bugs you mentioned in your 4 items. But have not changed the behaviour of the modal to apply changes even when dismissed without clicking OK.

I've made a JIRA card for that (EAGLE-1496)

M-Wicenec (approved):
works great!
editing should always apply on close imo, since all other modals behave like that. but otherwise perfect

PR #895, opened 2025-05-14
EAGLE-1484: Change font in codemirror

james-strauss-uwa:
Modify the global font change to use the CSS selector 'body' instead of '*'

## Summary by Sourcery

Apply global font-family via the body selector instead of the universal selector and clean up tooltip styling

Enhancements:
- Move global font-family declaration from the universal selector to the body selector
- Remove italic styling from <em> elements within tooltips


M-Wicenec (approved):
everything still looks right, great change

PR #897, opened 2025-05-16
EAGLE-1490: 'Select all' item to the palette triple dot menu

james-strauss-uwa:
Adds all items to the list of selected objects. Useful when dragging all components in to the graph to test them all at once.

## Summary by Sourcery

New Features:
- Add 'Select All' option to the palette triple-dot menu to select all components in a palette

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -2597,6 +2597,18 @@ export class Eagle {
         }
     }
 
+    selectAllInPalette = (palette: Palette): void => {
+        // close the palette menu
+        this.closePaletteMenus();
+
+        this.selectedObjects([]);
+        for (const node of palette.getNodes()){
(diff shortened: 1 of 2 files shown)


M-Wicenec (approved):
works, just an idea. but this could also be implemented by shift + left clicking the header.
i also just noticed that, besides the triple dot menu, you can right click the palette header. this presents mostly the same options except that this list is outdated. not sure what should be done there..

james-strauss-uwa:
Interesting, I had completely forgotten about the right click menu on palette headers.

It is probably less discoverable than the triple-dot menu. So i guess my thought would be to remove the right-click menu and make the triple-dot menu the only menu.

There is a least one right-click menu items that doesn't appear in the triple-dot menu, so we should move that over.

In any case, I'll create a JIRA issue for it: EAGLE-1494

PR #898, opened 2025-05-16
EAGLE-1177: Copy port description when renaming ports on Data components

james-strauss-uwa:
This JIRA issue mentioned a number of things, some of which had already been fixed in other issues. The remaining things were:

- When a user drags an edge between two Application nodes, and an intermediate Data node is created, we now copy the description as well as the name of the Application port.
- When a user drags an edge into empty space, and then generates a Data node, we now copy the description as well as the name of the Application port.

## Summary by Sourcery

Copy port description alongside port name when renaming Data component ports and creating new Data nodes during drag-and-drop operations.

Bug Fixes:
- Preserve port description when renaming Data ports after connecting Application and Data nodes
- Include port description when generating new Data nodes on drag-and-drop into empty space

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -4420,7 +4420,7 @@ export class Eagle {
             newNode.removeAllOutputPorts();
 
             // add InputOutput port for dataType
-            const newInputOutputPort = new Field(Utils.generateFieldId(), srcPort.getDisplayText(), "", "", "", false, srcPort.getType(), false, [], false, Daliuge.FieldType.Application, Daliuge.FieldUsage.InputOutput);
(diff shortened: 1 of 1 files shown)


M-Wicenec (changes requested):
i like this change, but i think it needs to be smarter/ more deliberate. Picture this scenario

1. if you have a copy app a file node and a random array app (not connected) 
2. you draw the first connection from copy app -> file node
3. the file input+output field's name and description change to that of the copy app's output port
4. you draw the connection from file -> random array
5. the files input+output port name and description get overwritten again to that of the random array app. 

this of course will always just show the name and description of the last connection made (since most data nodes use input+output ports by default)

does it make sense in this case to prioritise the name and description of the incoming app connection side for the name and description of the data input/output port field?

james-strauss-uwa:
After discussing this with @M-Wicenec , we decided to create a new ticket for his suggestion (EAGLE-1497), and to merge this work as it is.

PR #899, opened 2025-05-20
EAGLE-1493: Old graphs break EAGLE

james-strauss-uwa:
EAGLE is now able to load that directory of old graphs.

Error was caused by:
- node JSON using "oid" attribute instead of "id". Added check for that
- node JSON using "group" attribute instead of "parentId". Added helper function for that.


james-strauss-uwa:
@sourcery-ai review

M-Wicenec (approved):
nice!

PR #900, opened 2025-05-20
EAGLE-1492: Check graph after field value changed

james-strauss-uwa:
Also add an autoFix function to the case where Memory has an input AND has pydata set. Function will turn an empty string pydata into 'None', but don't overwrite anything else.

## Summary by Sourcery

Add an auto-fix for empty pydata conflicts and recheck the graph after parameter changes

Enhancements:
- Attach an auto-fix via Errors.ShowFix to replace empty pydata values with the default
- Invoke eagle.checkGraph in ParameterTable after modifying a graph to revalidate it

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -3938,6 +3938,9 @@ export class Eagle {
 
         // update the type of the field
         field.setType(newType);
+
+        // re-check the graph
+        this.checkGraph();
     }
 
     changeNodeParent = async () => {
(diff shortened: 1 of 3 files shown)


M-Wicenec (approved):
this works nicely for value changes. were still missing coverage for a call on check graph when changing a field type via the table though.

as for the double input on memory nodes. i had asked Andreas about this the other day and apparently the translator will now ignore the input on the pydata if a port is connected, so it doesnt really matter if there is or isnt something in there as long as one of the two is used. Im not sure what the correct approach for this is, since the auto fix will then essentially overwrite potential data that was in the pydata of the memory field which may be handy for the user to keep when testing with connections, should they choose to go back to using pydata. 
i will confirm with andreas, but potentially we should just get rid of that check entirely if it truly doesnt matter to the translator/engine.

id say for now, just push it though as we can always get rid of it in the future if we need to.

james-strauss-uwa:
Good points.

I'll have a quick look at re-checking the graph on field type changes in this branch.

I made the autoFix function somewhat limited so that it will not overwrite any data the user has entered there. The fix function will only replace an empty string with "None".

But as you say, we expect this will change, and we'll need to update it.

PR #905, opened 2025-05-23
EAGLE-1499: Fields with 'Unknown' type generate warnings even when isEvent flag set

james-strauss-uwa:
If the event flag is set, it doesn't matter what 'type' the field has. When the field is written to JSON, the type will be replaced by "Event".

- Remove warning generated for field with Unknown type, if the field is an event
- In Parameter Table, disable the display of type for event fields, instead show `<Event>`

### DIFF
--- a/src/Field.ts
+++ b/src/Field.ts
@@ -810,8 +810,8 @@ export class Field {
         //checks for input ports
         if(field.isInputPort()){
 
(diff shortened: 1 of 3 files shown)


M-Wicenec (approved):
looks good to me

PR #906, opened 2025-05-23
EAGLE-1494: Combine palette header right click menu and palette triple-dot menu

james-strauss-uwa:
Removed right click menu from palette header and stuck with the triple-dot menu.

Moved one missing function "Exclude From Search" over to the triple-dot menu.
Fixed bug where palette searchExclude didn't work.

## Summary by Sourcery

Consolidate palette header context actions into the triple-dot menu and correct searchExclude behavior to ensure excluded palettes bypass search filtering.

Bug Fixes:
- Fix palette searchExclude toggle so excluded palettes correctly bypass search filtering.

Enhancements:
- Remove palette header right-click menu and consolidate palette actions under the triple-dot menu.
- Move the Exclude/Include From Search actions to the palette’s triple-dot menu with updated labels.
- Update palette template logic to respect the searchExclude flag when displaying nodes.

### DIFF
--- a/src/RightClick.ts
+++ b/src/RightClick.ts
@@ -23,19 +23,6 @@ export class RightClick {
         RightClick.edgeDropSrcIsInput = null;
     }
 
-    static rightClickReloadPalette() : void {
-        const eagle: Eagle = Eagle.getInstance();
-        let index = 0
-        const palettes = eagle.palettes()
-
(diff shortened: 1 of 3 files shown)


M-Wicenec (approved):
looks good, i dont remember ever using right click on the palette anyways. lets see if anyone complains about it being removed

PR #907, opened 2025-05-27
EAGLE-1502: Node setGroupStart/setGroupEnd functions should clone fields from Daliuge.ts

james-strauss-uwa:
- Modified Node.setGroupStart and Node.setGroupEnd to clone "group_start" and "group_end" fields from Daliuge.ts, rather than create new fields completely from scratch.
- Also create canonical examples of "persist" and "streaming" fields in Daliuge.ts
- Use the new "persist" and "streaming" fields within JSON parsing code, instead of creating fields from scratch
- Modified a lot of Field.ts to make setX() functions return the field itself, so that function calls can be chained
- Fixed bug where "Closes Loop" radio button in edge inspector was calling the wrong function, which didn't do everything required to set Node field values as well as the "closes loop" flag.

## Summary by Sourcery

Use canonical field templates and fluent setter chaining to unify field cloning and initialization across the codebase, and correct the Closes Loop toggle handler.

Bug Fixes:
- Fix the Closes Loop radio button in the edge inspector to call the correct toggleEdgeClosesLoop function

Enhancements:
- Make all Field.setX() methods return the field instance to enable fluent chaining
- Define canonical groupStart, groupEnd, persist, and streaming Field examples in Daliuge.ts
- Refactor Node.setGroupStart/setGroupEnd and JSON import logic to clone these canonical fields instead of recreating them
- Update Utils, Eagle, and ParameterTable code to leverage chained setter calls when cloning and initializing fields

### DIFF
--- a/src/Daliuge.ts
+++ b/src/Daliuge.ts
@@ -193,6 +193,9 @@ export namespace Daliuge {
     export const funcCodeField = new Field(null, FieldName.FUNC_CODE, "", "def func_name(args): return args", "Python function code", false, Daliuge.DataType.Python, false, [], false, Daliuge.FieldType.Component, FieldUsage.NoPort);
     export const funcNameField = new Field(null, FieldName.FUNC_NAME, "", "func_name", "Python function name", false, Daliuge.DataType.Python, false, [], false, Daliuge.FieldType.Component, FieldUsage.NoPort);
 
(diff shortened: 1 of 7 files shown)


M-Wicenec (approved):
nice changes 👍

PR #910, opened 2025-05-29
Eagle 1488

M-Wicenec:
- converted old jquery event bindings to knockout
To elaborate a little bit, i was questioning if i should replace jquery with ko, because the jquery binding requires quite a bit less code, where as ko needs data-bind event handlers to be added on all html objects. ultimately for the sake of code readability, i chose to convert it to ko as that is what you would expect. this makes it easier to understand troubleshoot in the future. 

- made a generic function for hiding bootstrap dropdown menus and applied it to all examples in the ui.
- added a small buffer to the hide function to make use of the menus less finicky

## Summary by Sourcery

Refactor dropdown hide behavior by replacing legacy jQuery hover handlers with a buffered Knockout event-driven method, update all templates and code to use the new closeDropdownMenu function, and add CSS to suppress focus outlines on dropdown controls.

Enhancements:
- Introduce a generic closeDropdownMenu method with a 400ms buffer to hide Bootstrap dropdowns.
- Replace direct jQuery mouseleave bindings with Knockout data-bind event handlers across HTML templates and main/Eagle code.
- Remove the deprecated closePaletteMenus function and obsolete jQuery handlers in favor of the new KO-based approach.
- Add CSS rules to eliminate focus outlines and box shadows on table dropdown buttons.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -2571,7 +2571,7 @@ export class Eagle {
 
     sortPalette = (palette: Palette): void => {
         // close the palette menu
-        this.closeDropdownMenus();
+        this.closeDropdownMenu();
 
         const preSortCopy = palette.getNodes().slice();
 
@@ -2588,7 +2588,7 @@ export class Eagle {
 
     selectAllInPalette = (palette: Palette): void => {
         // close the palette menu
-        this.closeDropdownMenus();
+        this.closeDropdownMenu();
 
         this.selectedObjects([]);
         for (const node of palette.getNodes()){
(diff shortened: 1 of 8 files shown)


james-strauss-uwa (changes requested):
Just the few changes suggested by sourcery. Otherwise, it looks good.

M-Wicenec:
a few improvements
- rewrote the mouse out to use a jquery event binding, using a unique new class as not all dropdowns use the bootstrap classes. 
- ive added the mouseout to dismiss feature to the default value dropdown and type dropdown in the parameter table as well. (I have not added it to the selects because they are not bootstrap dropdowns, so they behave completely differently) 
- i figured out why some dropdown menu options didnt close the dropdown menu. it was because they had a 'clickbubble=false' data bind, removing it fixed the issue and i dont see why it was there in the first place.

M-Wicenec:
ive had a browse and it seems there is no easy way to programmatically show and hide the options for select type elements. people seem to only do this by keeping track of the select element's state and handling both opening and closing. i dont want to do that, it seems overkill and prone to errors.

PR #917, opened 2025-06-13
RTD Update: Fix failing build with update .yaml

myxie:
# Problem

There was an update to the .readthedocs.yaml format that broke existing builds if they didn't have the necessary features. This means our documentation has not been built ~18months. 

# Solution

I've updated this so hopefully the documentation is built. I will also enable the `readthedocs` setting that builds PRs so we can confirm before we merge that it is successful.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -2,10 +2,32 @@
 
 version: 2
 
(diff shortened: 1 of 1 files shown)


james-strauss-uwa (approved):
Looks great, thanks Ryan.

PR #923, opened 2025-06-30
Eagle 1513 - check for graph description

M-Wicenec:
- new function for "graphIsInitiated" this checks if there is an actual graph loaded or created. returns false on eagle's initial load
- highlighting description and graph info buttons in a warning color if graph descriptions are missing
- changes some tooltip descriptions if the description is missing

## Summary by Sourcery

Add a check for whether a graph has been named or created and use it to enhance the UI and validation around missing graph descriptions.

Bug Fixes:
- Use isInitiated() to block saving of graphs without a name

Enhancements:
- Introduce isInitiated() to determine when a graph is user-created
- Show fallback “No ... available” messages in the short and detailed description tooltips
- Highlight the short, detailed, and info buttons in warning color if descriptions are missing
- Push warning issues in graph validity check when short or detailed descriptions are absent
- Prevent saving an unnamed graph by using isInitiated() instead of a name check

Chores:
- Adjust table CSS to remove hidden display and set header width


src/LogicalGraph.ts
```diff
@@ -918,6 +933,35 @@ export class LogicalGraph {
...
+        if (this.isInitiated() && this.fileInfo().detailedDescription === ""){
+            return EagleConfig.getColor('graphWarning')
+        }
+
+        return ""
+    }, this);
+
+    getGraphInfoBtnColor : ko.PureComputed<string> = ko.pureComputed(() => {
+        //this excludes graphs that have not been initiated by the user (eagle has an empty graph by default)
+        if (this.isInitiated() && this.fileInfo().detailedDescription === "" || this.fileInfo().name != "" && this.fileInfo().shortDescription === ""){
```
james-strauss-uwa, line 957: Maybe you could simplify the logic here. My suggestion might not be exactly what you intended, I'm not sure.

```suggestion
        if (this.isInitiated() && (this.fileInfo().detailedDescription === "" || this.fileInfo().shortDescription === "")){
```

james-strauss-uwa (changes requested):
Looks good, just one minor suggestion to consider.

PR #924, opened 2025-06-30
potential fix for custom dropdowns alignment for parameter fields table

M-Wicenec:
fix vertical alignment of custom drop downs in the parameter fields table.

## Summary by Sourcery

Bug Fixes:
- Reset margins and add explicit inset and transform rules to correct dropdown placement for both top and bottom alignments in the parameter fields table.


M-Wicenec:
fixed on my laptop screen, will need to double check on a bigger monitor

PR #927, opened 2025-07-14
Eagle 1529

M-Wicenec:
swapping construct selection behaviour single click, as well as click and drag, will now select the construct and its children.
double click or alt + left click will select only the clicked construct.
right clicking a construct will pull up the right click menu for just the construct

## Summary by Sourcery

Update GraphRenderer to implement new construct selection behaviors: single click and drag select a group and its children, double-click or Alt+click select only the targeted construct, and right-click opens the context menu for the individual construct.

Enhancements:
- Single-click and click-drag now select a construct together with its children
- Double-click and Alt+Left-Click now select only the clicked construct without its descendants
- Right-click on a construct opens the context menu without altering the current selection
- Refactor selection logic to prioritize double-click and Alt+click handlers and remove legacy selection branches

### DIFF
--- a/src/GraphRenderer.ts
+++ b/src/GraphRenderer.ts
@@ -1103,7 +1103,7 @@ export class GraphRenderer {
             }   
 
             //check for alt clicking, if so, add the target node and its children to the selection
-            else if(event.button != 2 && !event.altKey&&node.isGroup()){
+            else if(event.button != 2 && !event.altKey&&node.isGroup() && !eagle.objectIsSelected(node)){
                 GraphRenderer.selectNodeAndChildren(node,GraphRenderer.shiftSelect)
             }
 
--- a/src/GraphRenderer.ts
+++ b/src/GraphRenderer.ts
@@ -1097,21 +1097,26 @@ export class GraphRenderer {
         // select handlers
         if(node !== null && event.button != 1 && !event.shiftKey){
 
+            //double click has highest priority, select only this node if it is a contruct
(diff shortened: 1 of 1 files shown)


src/GraphRenderer.ts
```diff
@@ -1097,21 +1097,26 @@ export class GraphRenderer {
...
+            }   
+
+            //check for alt clicking, if so, add the target node and its children to the selection
+            else if(event.button != 2 && !event.altKey&&node.isGroup() && !eagle.objectIsSelected(node)){
+                GraphRenderer.selectNodeAndChildren(node,GraphRenderer.shiftSelect)
+            }
+
             // check if shift key is down, if so, add or remove selected node to/from current selection | keycode 2 is the middle mouse button
-            if (node !== null && event.shiftKey && !event.altKey){
+            else if (node !== null && event.shiftKey && event.altKey){
```
james-strauss-uwa, line 1111: I don't understand the check for altKey AND shiftKey here. Could you add a comment?

james-strauss-uwa (approved):
Works well. I just have one question about the code for which a comment might be helpful. Thanks.

src/GraphRenderer.ts
M-Wicenec: looking at it again this part did nothing, there was a duplicate in the code. shift click on a construct adds or removes the construct and its children from the selection. shift + alt click on a construct adds or removes just the construct, without children.

PR #933, opened 2025-07-18
graph becomes editable while modal is open when dragging modal

M-Wicenec:
ive added a system that removes the modal background and allows for graph interaction when the modal is open if the user drags the modal aside. 
this system as well as the modal position resets when closing the modal.
ive had to edit some functions so they still work if the selected target changes while the modal is open

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -4503,44 +4503,44 @@ export class Eagle {
 
     editNodeDescription = async (): Promise<void> => {
         const markdownEditingEnabled: boolean = Setting.findValue(Setting.MARKDOWN_EDITING_ENABLED);
-
+        const node = this.selectedNode();
         let nodeDescription: string;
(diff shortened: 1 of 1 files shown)


james-strauss-uwa (approved):
Works well.

Just one minor thing you could look at:

1. Load a graph
2. Select a node
3. Click the "description edit" pencil icon on a field in the Parameter Table
4. Observe the description edit modal appear
5. Move the modal and observe that the dark background disappears and EAGLE is interactable
6. Click the same "description edit" pencil icon as step 3.
7. Observe the modal disappears

The current behaviour is OK, and could be merged. But if it is easy to improve, you could try:
- If the pencil icon for the same field and attribute, keep the modal (or toggle it?)
- If the pencil icon is for a different field or attribute, hide the current modal and open a new modal for that new attribute?

PR #934, opened 2025-07-18
Bug - fixed resizing causing sorting in parameter table

M-Wicenec:
the resizing handle event was bubbling up to the parent causing the table to be sorted by that column when resizing.

note* the knockout click bubble requires the click event to be bound otherwise it doesnt do anything. Hence why ive added those empty click bindings

## Summary by Sourcery

Prevent column resize handle clicks from triggering column sorting by suppressing event bubbling on resizer elements

Bug Fixes:
- Prevent column resize handle clicks from triggering column sorting due to event bubbling

Enhancements:
- Add empty click handlers with clickBubble:false to resizer divs to stop click events from propagating


james-strauss-uwa (approved):
Works well

PR #939, opened 2025-07-30
EAGLE-1551: Graph loading fixes

james-strauss-uwa:
I think I found the main issue. There were actually a few different errors.

There is a helper function that examines node JSON to try to determine the parent of the node. It looks at the "group", "parent" and "parentId" attributes in that order, exiting as soon as it sees the first one.

At some stage I changed the graph loading code from always depending on the "parentId" attribute, to instead re-use this helper function. Hoping to support older graphs.

However, the updateKeysToIds function I mentioned when we spoke updates the "parentId" attribute of all nodes, but crucially it did NOT remove the "group" attribute. So the "group" attribute was still being used as the parent in some situations. This caused all sorts of problems.

## Summary by Sourcery

Fix graph loading errors by removing legacy attributes and adding null checks for missing parent nodes.

Bug Fixes:
- Guard against null parent nodes in findDepthById and GraphRenderer to avoid runtime errors when a parent is missing
- Ensure subject IDs are only updated when the subject field is defined and non-null

Enhancements:
- Remove obsolete node.key and node.group properties in GraphUpdater to prevent stale group attributes from affecting parent resolution

### DIFF
--- a/src/GraphRenderer.ts
+++ b/src/GraphRenderer.ts
@@ -1089,8 +1089,10 @@ export class GraphRenderer {
             //checking if the node is inside of a construct, if so, fetching it's parent
             if(node.getParentId() != null){
                 const parentNode = eagle.logicalGraph().findNodeByIdQuiet(node.getParentId())
-                $('#'+parentNode.getId()).removeClass('transition')
-                GraphRenderer.NodeParentRadiusPreDrag = parentNode.getRadius()
+                if (parentNode !== null){
+                    $('#'+parentNode.getId()).removeClass('transition')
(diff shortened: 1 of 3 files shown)


james-strauss-uwa:
I've also tested with Alex' dingo graphs. For example: http://localhost:8888/?service=GitHub&repository=alxndrwllmsn/EAGLE-graph-repo&branch=master&path=dingo/dstack&filename=dstack_uvgrid_image_stack.graph

Which all seem to work.

M-Wicenec (approved):
fixed it for me

PR #940, opened 2025-07-30
EAGLE-1552: Fix issue when writing empty string as integer to JSON

james-strauss-uwa:
Also fixed issue created as a result of the first issue. when expecting an value and reading null from JSON


M-Wicenec (approved):
this looks good

PR #941, opened 2025-08-07
EAGLE-1562: Error loading dingo palette

james-strauss-uwa:
http://localhost:8888/?service=GitHub&repository=ICRAR/EAGLE-graph-repo&branch=master&path=dingo&filename=dingo.palette

Fixed problem with missing construct nodes. 
Use better icon for 'Description' nodes

## Summary by Sourcery

Add null checks for missing construct nodes to prevent errors and update the description category to use a neutral icon

Bug Fixes:
- Guard against null construct nodes before accessing their properties to fix palette loading errors

Enhancements:
- Change the Description category to use the default 'icon-none' icon


M-Wicenec (approved):
This fixed the loading errors. I am getting some of the following warnings are they of any concern?
 "LogicalGraph.ts:583 findNodeById(): could not find node with id ( c04cebf3-5dab-4310-81c5-cecd14fe95ad )"
 
 i also noticed since a lot of the categories are old or unknown in the palette, we can actually pick a category for those in the inspector, which is good.

PR #943, opened 2025-08-08
EAGLE-882: Add undo snapshot after component update

james-strauss-uwa:
After component update:
- flag graph as modified
- recheck for correctness
- create undo snapshot

## Summary by Sourcery

After updating components, flag the graph as modified, recheck its correctness, and create an undo snapshot to capture the state.

Enhancements:
- Record an undo snapshot after component updates
- Flag the logical graph as modified and trigger value mutation
- Revalidate the graph for correctness after component updates


M-Wicenec (approved):
seems to work, a bit confusing since there is no feedback to the user at all as to what the update function is doing, but thats not a problem on this branch. we'll have to address the update feature at some point.

PR #944, opened 2025-08-08
EAGLE-1038: Allow users to update a single component

james-strauss-uwa:
Add right click menu item 'Update' on graph nodes, that updates the components


M-Wicenec (approved):
works nicely

PR #953, opened 2025-08-19
EAGLE-1572: Obey SHOW_GRAPH_WARNINGS setting when colouring graph edges

james-strauss-uwa:
Fixed bug where edge warnings and errors would be shown even when the setting to show them was disabled.


M-Wicenec (approved):
👍

PR #957, opened 2025-08-21
reducing scroll sensitivity of types dropdown in parameter table

M-Wicenec:
created a function which we can trigger on scrollable elements that need to scroll slower.

## Summary by Sourcery

Add reduced-scroll sensitivity to type-selection dropdowns in the node parameter table by introducing a slowScroll handler and binding it to mousewheel events; refine modal event handler signatures and prevent unwanted click propagation in dropdown menus.

Enhancements:
- Introduce slowScroll method in Eagle to halve mousewheel scroll delta on scrollable elements
- Bind slowScroll to mousewheel events on parameter table type dropdown menus for smoother vertical scrolling
- Disable click event bubbling on type dropdown menus to avoid unintended interactions
- Add explicit event parameter typings to Bootstrap modal shown and header mousedown handlers

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -62,7 +62,6 @@ import { Utils } from './Utils';
 import { GraphUpdater } from "./GraphUpdater";
 import { GraphConfigurationsTable } from "./GraphConfigurationsTable";
 import { versions } from "./Versions";
-import { event } from "jquery";
 
 
 export class Eagle {
--- a/templates/node_parameter_table.html
+++ b/templates/node_parameter_table.html
@@ -370,7 +370,7 @@
                                                                         arrow_drop_down
                                                                     </span>
                                                                 </button>
-                                                                <ul class="dropdown-menu dropdown-menu-end dropdown-area" data-bind="event:{mousewheel: $root.slowScroll},attr:{id:'typeFor_'+$data.getId()},clickBubble:false">
+                                                                <ul class="dropdown-menu dropdown-menu-end dropdown-area" data-bind="event:{mousewheel: $root.slowScroll}, attr:{id:'typeFor_'+$data.getId()}">
                                                                     <!-- ko foreach:$root.types() -->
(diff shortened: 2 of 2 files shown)


src/Eagle.ts
```diff
@@ -62,6 +62,7 @@ import { Utils } from './Utils';
 import { GraphUpdater } from "./GraphUpdater";
 import { GraphConfigurationsTable } from "./GraphConfigurationsTable";
 import { versions } from "./Versions";
+import { event } from "jquery";
```
james-strauss-uwa, line 65: Do we need this, or was it added automatically by the editor?

templates/node_parameter_table.html
```diff
@@ -321,7 +321,7 @@ <h5 class="bottomWindowHeader">Node Fields Table: <span>(No selection)</span></h
                                                         <!-- ko if: Setting.findValue(Setting.ALLOW_COMPONENT_EDITING) -->
                                                             <div class="dropdown parameterTableTypeCustomSelect">
                                                                 <a class="dropdown-toggle dropdown-control" data-bind="text:$data.defaultValue, disabled: ParameterTable.getNodeLockedState($data)" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false"></a>
-                                                                <div class="dropdown-menu dropdown-menu-right dropdown-area" aria-labelledby="parameterTableTpeCustomSelect">
+                                                                <div class="dropdown-menu dropdown-menu-right dropdown-area" aria-labelledby="parameterTableTpeCustomSelect" data-bind="event:{mousewheel: $root.slowScroll}">
```
james-strauss-uwa, line 324: The two changes in this file are slightly different. One adds "clickBubble:false" and the other does not.

Should it be added to both? Or not required in either? Or is there a reason for having one and not the other?

james-strauss-uwa (changes requested):
Just two minor issues that I noticed when looking at the code.

If you're happy that both the changes are intentional, then consider this approved.

templates/node_parameter_table.html
M-Wicenec: leftover from testing an issue i had, no longer needed.

PR #960, opened 2025-09-01
Eagle 1575 - field value type UI improvements

M-Wicenec:
removed redundant "getHTMLType" function
created "isBooleanTypeValue" etc functions to use in the knockout to show the correct UI element
added support for standard python types (float vs Float, bool vs Boolean, etc.) the UI will now work correctly for both.
some related code clean up for parameter table, config parameter table, inspector and edit field modal

## Summary by Sourcery

Refactor field input type logic to use dedicated type-check functions, support Python type aliases, and update UI templates accordingly for consistent rendering of parameter fields.

New Features:
- Support Python built-in type names (int, float, bool) for field value rendering

Enhancements:
- Introduce isFloatValueType, isIntegerValueType, isBooleanValueType, isSelectValueType, and isStringValueType methods on Field and remove redundant getHtmlInputType
- Add lowercase python type aliases (int, float, bool) to the Daliuge.FieldName enum
- Revamp knockout templates for config and node parameter tables, edit-field modal, and inspector to use new type-check bindings and simplify conditional rendering

Chores:
- Clean up outdated code and redundant markup in parameter tables, modals, and inspector templates

### DIFF
--- a/src/Daliuge.ts
+++ b/src/Daliuge.ts
@@ -102,7 +102,6 @@ export namespace Daliuge {
         Boolean = "Boolean",
         bool = "bool",
         Select = "Select",
-        Password = "Password",
         Json = "Json",
         Python = "Python"
     }
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -232,7 +232,6 @@ export class Eagle {
             Daliuge.DataType.Integer,
             Daliuge.DataType.Json,
             Daliuge.DataType.Object,
-            Daliuge.DataType.Password,
             Daliuge.DataType.Python,
             Daliuge.DataType.Select,
             Daliuge.DataType.String


M-Wicenec:
> One of the sourcery ai comments says that these changes remove the specific handlers for JSON and Password fields. This isn't 100% correct, but did make me think.
> 
> The default "stringValueType" will handle JSON well. And clicking the "pencil" icon to edit the value will run `ParameterTable.requestEditFieldValue()`. That function will call `ParameterTable.isCodeField()` to determine whether the editor modal defaults to editing plain text or code. All good.

yes youre right, its still just a string, in the table displayed in a textbox. but the edit button will sus out the editing type.

> Interestingly, the `ParameterTable.isCodeField()` function only looks at the NAME of the field, if it is "func_code" or "command", then the value will be edited as code instead of plain text. I might make a ticket to change this. I think that `isCodeField()` should return true for fields with type === JSON too.

i think this makes sense, did you have to specify the language when initiating the code editor? maybe we could also have a 'python' type or something.

> However, I think we do lose the handler for the "password" field type. Previously, the code would set the "type" of the `<input>` to "password", and the text would therefore be hidden.

i just cross checked master as well, we did not handle passwords any different before, they were just shown as strings. hence why i did not pay them any attention in this ticket

> I don't think the "password" field type is used much. Not even in our existing Docker, NGAS or S3 components. So I think we should remove the "password" type (remove Daliuge.DataType.Json).
> 
> Otherwise looks great.

you wanted me to remove the password field type in this branch?

james-strauss-uwa:
Yes, please remove Daliuge.DataType.Password in this branch.

Sorry, somehow I wrote "Daliuge.DataType.Json" in my previous message!

Regarding the code editor: Yes, you need to specify the language when opening the code editor. In `Utils.requestUserCode()` , the first parameter is "json", "python" or "text".

M-Wicenec:
ive removed the password type and a related function.

james-strauss-uwa (approved):
Thanks, looks great 👍

PR #962, opened 2025-09-05
EAGLE-1563: Save and load GraphConfig files

james-strauss-uwa:
Allows users to save and load GraphConfig files.

Saving is via:
- an "action" (save icon) in the GraphConfigurationsTable
- the "Save" item in the Config menu

Loading is done by:
- clicking a GraphConfig in the Repositories tab
- the "Load" item in the Config menu

The main change to the code is the new FileInfo block inside each GraphConfig. Corresponding changes also made to the V4 schema, and to the sample V4 file found in the e2e/data directory.

New class called FileLocation, to keep the file service, repo, branch, path, filename, hash and url all together.

Limitations:

- If the user opens a GraphConfig file that doesn't belong to the open LogicalGraph, then EAGLE will show an error message. Eventually we want it to close the current LG and open the LG to which the GraphConfig belongs (EAGLE-1586).
- No checking whether a GraphConfig occurs multiple times in a LogicalGraph
- Individual GraphConfig files are not checked whether they validate a schema (EAGLE-1588)

## Summary by Sourcery

Enable saving and loading of GraphConfig files and refactor repository metadata handling using a new FileLocation abstraction

New Features:
- Allow saving and loading of GraphConfig files locally and via Git through UI actions in the configurations table, Config menu, and repository file selector
- Define a new GraphConfig file type with end-to-end support for serialization, deserialization, JSON schema validation, and custom file extension handling

Enhancements:
- Introduce FileLocation class to centralize repository metadata and refactor FileInfo to use FileLocation for location and graphLocation fields
- Refactor GraphConfig to include FileInfo, migrate legacy attributes into modelData, and update LogicalGraph.addGraphConfig to record parent graph location, set active config, and open the table with undo support
- Update UI templates and modals to handle GraphConfig inputs, menu items, table rows, and validation for the .graphConfig extension

Documentation:
- Update V4 JSON schema and e2e sample files to include GraphConfig support

### DIFF
--- a/templates/modals.html
+++ b/templates/modals.html
@@ -224,9 +224,6 @@
                     <div class="row">
                         <p id="optionsModalMessage">Message</p>
                     </div>
-                    <div class="row" id="confirmModalDontShowAgain">
-                        <p>Don't ask me again <button class="material-symbols-outlined">check_box_outline_blank</button></p>
-                    </div>
                 </div>
             </div>
             <div class="modal-footer">


james-strauss-uwa:
After talking to @M-Wicenec there are several things to fix here:

- Add a user warning before saving a GraphConfig where the modelData location does not match the LogicalGraph modelData location. Suggest the user run "fix all".
- When loading a GraphConfig, and the config matches the LogicalGraph, check if the id of the GraphConfig already exists. If so, offer the user options: Overwrite existing config, add as new config, cancel.
- When loading a GraphConfig, and the config does NOT match, offer the user options: OK, Add anyway.
- Maybe also add information to the config mismatch dialog that displays how many of the config's nodes and ports can be found in the LogicalGraph.

M-Wicenec (commented):
looks good.
only minor thing ive still found is that when loading configs, you get that pop up asking if you want to overwrite or rename. On this pop up you get the generic "don't ask me again" checkbox. 
I suppose it doesnt really make sense to ask this here, and it also doesnt seem to work

PR #966, opened 2025-09-19
EAGLE-1592: Fixed broken 'edit description' buttons in inspector

james-strauss-uwa:
Fixed bug where 'edit description' buttons in inspector were broken


M-Wicenec (approved):
👍

PR #969, opened 2025-09-23
EAGLE-1587: GlobalVariable component

james-strauss-uwa:
Moved EnviromentVariables component to Legacy categoryType
Added a new "Global" categoryType
Added GlobalVariable component (with the Global categoryType)
Removed unused Category.sortOrder (everything is shown alphabetically now)
Added map of Legacy components categories to recommended upgrades. The fix functions now update to the recommended category by default
Fix functions now update categoryType as well as category. To handle cases where the recommended update is of a different categoryType
Don't warn user if GlobalVariable component is completely disconnected from the graph
Updated CI test graph
Modified category colours to use the same (hex) format

One way to test this is to:
- Open the graph: http://localhost:8888/?service=GitHub&repository=ICRAR/EAGLE_test_repo&branch=master&path=eagle_test_graphs/daliuge_tests/translator/logical_graphs&filename=ArrayLoop_Global.graph
- Press 'f' to fix all issues
- Observe the EnvironmentVariables component changed to a GlobalVariable component

## Summary by Sourcery

Introduce a new GlobalVariable category, retire EnvironmentVariables into legacy, streamline category ordering, and provide automated upgrade paths with fix actions while updating related tests.

New Features:
- Add GlobalVariable component as a new Construct category

Enhancements:
- Move EnvironmentVariables to legacy and map it to upgrade by default
- Remove Category.sortOrder and deprecated SortOrder enum to present categories alphabetically
- Add LEGACY_CATEGORIES_UPGRADES map and Utils.getLegacyCategoryUpdate for automated legacy category migrations
- Extend fix actions to update both category and categoryType when migrating nodes
- Suppress disconnected-node warnings for GlobalVariable components

Tests:
- Refresh CI test graph to accommodate new category changes


myxie (approved):

M-Wicenec (approved):
looks good!
Im assuming it has been decided that the rendering of the global environment nodes in the graph should be like an application rather than a data node?

james-strauss-uwa:
> looks good! Im assuming it has been decided that the rendering of the global environment nodes in the graph should be like an application rather than a data node?

I actually did forget to update Utils.getRadiusForNode() to have a case for the new "Global" categoryType.

Do you think it would look better the same size as Data, or the default size it is now (same as Applications)?

M-Wicenec:
> > looks good! Im assuming it has been decided that the rendering of the global environment nodes in the graph should be like an application rather than a data node?
> 
> I actually did forget to update Utils.getRadiusForNode() to have a case for the new "Global" categoryType.
> 
> Do you think it would look better the same size as Data, or the default size it is now (same as Applications)?

last I spoke to @awicenec we thought it should like a data node in the graph. But perhaps we should ask again.
The differences are the radius and the label location being on the bottom rather than the at the top. data node labels can now also be hidden using the new function from the navbar. I wonder if it were enough for the node.isData() function to be updated?

james-strauss-uwa:
> last I spoke to @awicenec we thought it should like a data node in the graph. But perhaps we should ask again. The differences are the radius and the label location being on the bottom rather than the at the top. data node labels can now also be hidden using the new function from the navbar. I wonder if it were enough for the node.isData() function to be updated?

I agree, we should render it like a Data node. I've updated Utils.getRadiusForNode() to use the smaller radius. Also, I ended up defining a new Node.isGlobal() and using that in addition to Node.isData(), instead of modifying Node.isData() to include Global components.

So now it uses the smaller radius, label on the bottom, and gets hidden along with Data nodes when that setting is ON.

Could you check it one more time? Thanks.

M-Wicenec:
thanks, looks good now!

PR #970, opened 2025-09-23
EAGLE-1593: Restructure Utils.requestUserConfirm

james-strauss-uwa:
Improved the handing of this modal to prevent an error from being thrown when the user answered negative to the confirmation dialog.

The behaviour has slightly changed. Processing of the user's response is no longer triggered by pressing the affirmative button, but instead when the dialog is hidden. So a slightly longer delay is required in the unit tests (to account for the length of time taken for the modal dialog to animate to its hidden state).

## Summary by Sourcery

Restructure the user confirmation flow so Utils.requestUserConfirm returns a boolean and only executes actions when the user affirms, preventing errors on negative responses and deferring callbacks until the modal hides.

Enhancements:
- Change Utils.requestUserConfirm signature to return a boolean and reject only on user cancellation
- Refactor confirmation modal to set completion and confirmation flags on button clicks and invoke callbacks on hidden event
- Update Eagle and Repositories code to capture the confirmed flag and conditionally execute load, delete, reload, close, and category-change operations
- Maintain focus on the affirmative button when the modal is shown

Tests:
- Add waitForTimeout delays in e2e tests to account for modal hide animation

### DIFF
--- a/e2e/addingAndRemovingRepositories.spec.ts
+++ b/e2e/addingAndRemovingRepositories.spec.ts
@@ -38,6 +38,7 @@ test('Adding and Removing Repositories', async ({ page }) => {
   //removing the repo
   await page.locator('.repoContainer').filter({has:page.locator(repoHTMLId)}).hover()
   await page.locator(repoHTMLId + "-eject").click()
+  await page.waitForTimeout(500);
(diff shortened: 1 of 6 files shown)


M-Wicenec (approved):
👍

PR #971, opened 2025-09-26
eagle-1594 -  improve readability of graph edge flow

M-Wicenec:
- adjusted edge thickness
- adjusted edge arrow size
- adjusted edge length before arrow is hidden
- adjusted edge length before edge becomes straight
- added support for the previous two attributes (edge lengths) to be controlled separately
- added edge thickness to the eagle config

## Summary by Sourcery

Enhance edge rendering by making thickness and arrow display parameters configurable, refining threshold defaults, and streamlining arrow drawing logic for both curved and straight edges

New Features:
- Introduce EDGE_THICKNESS config with CSS integration for controlling edge stroke width
- Add separate EDGE_DISTANCE_ARROW_VISIBILITY and STRAIGHT_EDGE_SWITCH_DISTANCE thresholds for arrow visibility and edge curvature

Enhancements:
- Adjust default values of EDGE_ARROW_SIZE, EDGE_DISTANCE_ARROW_VISIBILITY and STRAIGHT_EDGE_SWITCH_DISTANCE to improve graph readability
- Refactor GraphRenderer to decouple arrow visibility and curvature logic, computing arrow position and angle for both curved and straight edges

### DIFF
--- a/src/GraphRenderer.ts
+++ b/src/GraphRenderer.ts
@@ -796,10 +796,6 @@ export class GraphRenderer {
         const isShortEdge: boolean = edgeLength < EagleConfig.STRAIGHT_EDGE_SWITCH_DISTANCE;
         const hideArrow: boolean = edgeLength < EagleConfig.EDGE_DISTANCE_ARROW_VISIBILITY;
 
-        if (edge !== null){
-            edge.setIsShortEdge(hideArrow)
-        }
-
         
         // -------------generate bezier curve control points---------------
         
--- a/src/Edge.ts
+++ b/src/Edge.ts
@@ -162,22 +162,6 @@ export class Edge {
         this.selectionRelative = !this.selectionRelative;
     }
 
-    setIsShortEdge = (value:boolean) : void => {
-        this.isShortEdge(value)
-    }
-
-    getIsShortEdge = () : boolean => {
-        return this.isShortEdge()
-    }
-
-    getArrowVisibility = () : string => {
-        if (this.isShortEdge()){
-            return 'hidden' 
-        }else{
-            return 'visible'
-        }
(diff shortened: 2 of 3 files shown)


src/GraphRenderer.ts
```diff
@@ -794,9 +794,10 @@ export class GraphRenderer {
 
         //determining if the edge's length is below a certain threshold. if it is we will draw the edge straight and remove the arrow
         const isShortEdge: boolean = edgeLength < EagleConfig.STRAIGHT_EDGE_SWITCH_DISTANCE;
+        const hideArrow: boolean = edgeLength < EagleConfig.EDGE_DISTANCE_ARROW_VISIBILITY;
 
         if (edge !== null){
-            edge.setIsShortEdge(isShortEdge)
+            edge.setIsShortEdge(hideArrow)
```
james-strauss-uwa, line 800: I guess we should probably change the name of the setIsShortEdge() function here.

If you are calculating whether to hide the arrow, and storing that value on the edge, then "isShortEdge" is not a good name. Change it to 'hideArrow' or something similar.

Or do we even need to store the value of hideArrow on the edge? Perhaps we can remove the "isShortEdge" variable in the Edge class (and the get/set functions? It looks like we only use it locally within this function?

james-strauss-uwa (changes requested):
If it's possible to remove the Edge.isShortEdge variable, we should do that. Thanks.

Otherwise, looks good.

src/GraphRenderer.ts
M-Wicenec: i was able to remove it and handle arrow visibility within the edge drawing function without the need for a ko data bind

PR #975, opened 2025-10-06
EAGLE-1582: Modified RepositoryCommit class to contain a FileLocation

james-strauss-uwa:
This is primarily a simplification of the RepositoryCommit class.

Replacing many variables (repositoryService, repositoryName, repositoryBranch, filePath, fileName) with a single FileLocation structure.

This uncovered a type-mismatch in the way the "git commit" modal returned data. So I defined a new set of types (in Modals.ts) and modified the callback and callback handlers to use those type definitions. So it should be more difficult for the modals and callback functions to get out-of-sync in the future.

### DIFF
--- a/src/Eagle.ts
+++ b/src/Eagle.ts
@@ -2031,9 +2031,9 @@ export class Eagle {
             }
 
             // check repository name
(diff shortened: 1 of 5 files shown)


M-Wicenec (approved):
looks good,
just very minor comment regarding spacing around ':' when defining constants

PR #994, opened 2025-12-15
EAGLE-1522: set 'strictNullChecks' flag in Typescript config

james-strauss-uwa:
Set the 'strictNullChecks' flag to 'true' in the Typescript config

This required changes to almost all Typescript files, to make sure we handle all cases where variables are null or undefined.

Significant changes:

- Node default radii moved to CategoryData
- The Field constructor can't be given a null value for the node parameter. So a Field can't exist without belonging to a Node. Also, a Field now can't have a null id.
- `Setting.findValue()` is now a generic, so we can specify the type we expect to be returned. For example: `Setting.findValue<string>(Setting.DOCKER_HUB_USERNAME, "")` will return a string. Also notice that we now specify a default value, in case the Setting is undefined.
- `Setting.find(X).setValue(y)` can now be done with `Setting.setValue(X, y)`
- `Translator.DEFAULT_TRANSLATION_ALGORITHM` is now defined `"alg-1"`.
- There are still some consistency issues, but some functions (e.g. `getParent()`) return null if nothing is set. Some functions (e.g. `findNodeById()`) return undefined is the object can't be found. This could still use improvement.
- `Node.clear()`, `Edge.clear()` and `Field.clear()` were unused and removed
- Defined `GraphRenderer.position2d` type to use everywhere we use {x:number, y:number}
- Defined `Setting.validValueTypes` type (string | number | boolean) that specifies the type that may be stored in Settings


src/Eagle.ts
```diff
@@ -503,12 +535,17 @@ export class Eagle {
...
-            bottomWindow = $('#bottomWindow').height()
+        if(Setting.findValue<boolean>(Setting.BOTTOM_WINDOW_VISIBLE, false)){
+            const bottomWindowHeight = $('#bottomWindow').height()
+            bottomWindow = bottomWindowHeight !== undefined ? bottomWindowHeight : 0
         }
 
+        // get width and height of the logical graph parent container
+        const logicalGraphParentWidth = $('#logicalGraphParent').width() || 0;
+        const logicalGraphParentHeight = $('#logicalGraphParent').height() || 0;
+
```
M-Wicenec, line 546: if this defaults to 0 we should probably spit out a console warning, so we know that there is an issue

src/Eagle.ts
```diff
@@ -879,7 +924,7 @@ export class Eagle {
             const reader = new FileReader();
             reader.readAsText(file, "UTF-8");
             reader.onload = function (evt) {
-                const data: string = evt.target.result.toString();
+                const data: string = evt.target?.result?.toString() || "";
```
M-Wicenec: we might need another catch here? does it cause issues if we default to ""

src/Eagle.ts
```diff
@@ -1315,7 +1389,7 @@ export class Eagle {
...
```
M-Wicenec: again probably need a catch for the empty string case

src/Eagle.ts
```diff
@@ -1386,15 +1465,14 @@ export class Eagle {
...
```
M-Wicenec: one more

src/Eagle.ts
```diff
@@ -4553,19 +4841,28 @@ export class Eagle {
...
         
         while (!suitablePositionFound && numIterations <= MAX_ITERATIONS){
+            const leftWindowVisible = Setting.findValue<boolean>(Setting.LEFT_WINDOW_VISIBLE, false);
+            const rightWindowVisible = Setting.findValue<boolean>(Setting.RIGHT_WINDOW_VISIBLE, false);
+            const bottomWindowVisible = Setting.findValue<boolean>(Setting.BOTTOM_WINDOW_VISIBLE, false);
+
+            // get logical graph display area dimensions
+            const logicalGraphParentWidth = $('#logicalGraphParent').width() || 0;
+            const logicalGraphParentHeight = $('#logicalGraphParent').height() || 0;
+            const bottomWindowHeight = $('#bottomWindow').height() || 0;
```
M-Wicenec: may need a catch for if it defaults to 0 for error handling

src/GraphConfig.ts
```diff
@@ -321,19 +333,19 @@ export class GraphConfigNode {
...
+        this.value = ko.observable<string | null>(null);
         this.comment = ko.observable("");
     }
 
     clone = (): GraphConfigField => {
-        const result = new GraphConfigField();
+        const result = new GraphConfigField(this.field());
 
-        result.field(this.field());
+        //result.field(this.field());
```
M-Wicenec: no longer needed?

src/GraphRenderer.ts
```diff
@@ -143,20 +149,12 @@ ko.bindingHandlers.graphRendererPortPosition = {
         //this handler is for a PORT position, meaning it will run twice for a field that has both input and output ports
         //the update function is called initially and then whenever a change to a utilised observable occurs
         const eagle : Eagle = Eagle.getInstance();
-        const n: Node = ko.utils.unwrapObservable(valueAccessor()).n;
+        //const n: Node = ko.utils.unwrapObservable(valueAccessor()).n;
```
M-Wicenec: not needed?

src/GraphRenderer.ts
```diff
@@ -236,6 +234,7 @@ ko.bindingHandlers.graphRendererPortPosition = {
                     break;
                 default:
                     console.warn("disconnected field with dataType:", dataType);
+                    averageAngle = 0
```
M-Wicenec: if you set the default value to 0 in the definition you wont need to set it here

src/GraphRenderer.ts
```diff
@@ -911,13 +915,14 @@ export class GraphRenderer {
...
-        if(edge != null || addArrowForce){
-            let arrowContainer
+        if(edge !== null || addArrowForce){
+            let arrowContainer = $('#draggingEdge polygon')
 
-            if(addArrowForce){
-                arrowContainer = $('#draggingEdge polygon')
-            }else{
-                arrowContainer = $('#'+edge.getId() +" polygon")
+            // TODO: not sure about this change, it think the structure of branches here is a little hard to understand
```
M-Wicenec: on review, this seems fine. you can get rid of this comment

src/GraphRenderer.ts
```diff
@@ -1515,11 +1548,15 @@ export class GraphRenderer {
...
+            GraphRenderer.resizeConstruct(parent)
+        }
 
         //updating the parent construct's "pre-drag" size at the end of parenting all the nodes
-        // TODO: check this line, could it be: GraphRenderer.nodeParentRadiusPreDrag = parent.getRadius()
-        GraphRenderer.nodeParentRadiusPreDrag = Eagle.getInstance().logicalGraph().getNodeById(parent?.getId())?.getRadius()
+        if (parent !== null){
+            GraphRenderer.nodeParentRadiusPreDrag = parent.getRadius();
+        }
+        //GraphRenderer.nodeParentRadiusPreDrag = Eagle.getInstance().logicalGraph().getNodeById(parent?.getId())?.getRadius()
```
M-Wicenec: can get rid of this?

src/GraphRenderer.ts
```diff
@@ -2189,16 +2249,25 @@ export class GraphRenderer {
 
     static SCREEN_TO_GRAPH_POSITION_X(x:number) : number {
         const eagle = Eagle.getInstance();
-        if(x===null && GraphRenderer.dragCurrentPosition){
-            x = GraphRenderer.dragCurrentPosition.x
+        if(x===null){
+            if (GraphRenderer.dragCurrentPosition !== null){
+                x = GraphRenderer.dragCurrentPosition.x
+            } else {
+                x = 0;
```
M-Wicenec, line 2236: this should probably have a warning to prevent errors without trace

M-Wicenec: also may not be necessary because it is handed to the function from outside

src/GraphRenderer.ts
```diff
@@ -2189,16 +2249,25 @@ export class GraphRenderer {
...
-    static SCREEN_TO_GRAPH_POSITION_Y(y:number) : number {
+    static SCREEN_TO_GRAPH_POSITION_Y(y:number | null) : number {
         const eagle = Eagle.getInstance();
-        if(y===null && GraphRenderer.dragCurrentPosition){
-            y = GraphRenderer.dragCurrentPosition.y
+        if(y===null){
+            if (GraphRenderer.dragCurrentPosition !== null){
+                y = GraphRenderer.dragCurrentPosition.y
+            } else {
+                y = 0;
```
M-Wicenec, line 2250: same as the x case above

src/LogicalGraph.ts
```diff
@@ -409,15 +434,23 @@ export class LogicalGraph {
             // if source node or destination node is a construct, then something is wrong, constructs should not have ports
             if (sourceNode.getCategoryType() === Category.Type.Construct){
                 const srcIdAndPort = sourceNode.findPortInApplicationsById(edge.getSrcPort().getId());
-                const warning = "Updated source node of edge " + edge.getId() + " from construct " + edge.getSrcNode().getId() + " to embedded application " + srcIdAndPort.node.getId();
-                errorsWarnings.warnings.push(Errors.Message(warning));
-                edge.getSrcNode().setId(srcIdAndPort.node.getId());
+                if (typeof srcIdAndPort.node === 'undefined'){
+                    // TODO: add error
```
M-Wicenec: add error?

src/LogicalGraph.ts
```diff
@@ -409,15 +434,23 @@ export class LogicalGraph {
...
+                    edge.getSrcNode().setId(srcIdAndPort.node.getId());
+                }
             }
             if (destinationNode.getCategoryType() === Category.Type.Construct){
                 const destKeyAndPort = destinationNode.findPortInApplicationsById(edge.getDestPort().getId());
-                const warning = "Updated destination node of edge " + edge.getId() + " from construct " + edge.getDestNode().getId() + " to embedded application " + destKeyAndPort.node.getId();
-                errorsWarnings.warnings.push(Errors.Message(warning));
-                edge.getDestNode().setId(destKeyAndPort.node.getId());
+                if (typeof destKeyAndPort.node === 'undefined'){
+                    // TODO: add error
```
M-Wicenec: this one too?

src/Node.ts
```diff
@@ -75,7 +75,7 @@ export class Node {
                                                    // TODO: unused? shall we remove it?
 
     constructor(name : string, description : string, comment : string, category : Category){
-        this.id = ko.observable(Utils.generateNodeId());
+        this.id = ko.observable("" as NodeId);//Utils.generateNodeId());
```
M-Wicenec: comment here, also is this correct?

src/Node.ts
```diff
@@ -1601,16 +1603,14 @@ export class Node {
             if (categoryData.categoryType !== Category.Type.Construct){
                 errorsWarnings.errors.push(Errors.Message("Attempt to add inputApplication to unsuitable node: " + category));
             } else {
-                node.inputApplication(Node.fromOJSJson(nodeData.inputApplication, errorsWarnings, isPaletteNode));
-                node.inputApplication().setEmbed(node);
```
M-Wicenec, line 1605: this line seems to have been lost, do we need this?

src/Node.ts
```diff
@@ -1601,16 +1603,14 @@ export class Node {
...
+                node.setInputApplication(Node.fromOJSJson(nodeData.inputApplication, errorsWarnings, isPaletteNode));
             }
         }
         if (typeof nodeData.outputApplication !== 'undefined' && nodeData.outputApplication !== null){
             if (categoryData.categoryType !== Category.Type.Construct){
                 errorsWarnings.errors.push(Errors.Message("Attempt to add outputApplication to unsuitable node: " + category));
             } else {
-                node.outputApplication(Node.fromOJSJson(nodeData.outputApplication, errorsWarnings, isPaletteNode));
-                node.outputApplication().setEmbed(node);
```
M-Wicenec, line 1613: here as well

src/RightClick.ts
```diff
@@ -442,10 +461,11 @@ export class RightClick {
 
     // TODO: event var used in function is the deprecated global, we should get access to the event via some other method
     static edgeDropCreateNode = (data: Node[]) : void => {
-        RightClick.requestCustomContextMenu(data, 'edgeDropCreate')
+        RightClick.requestCustomContextMenu(data, 'edgeDropCreate');
 
         // prevent bubbling events
-        event.stopPropagation();
+        // TODO: get event from somewhere instead of global
```
M-Wicenec: i believe the event comes from the ko event trigger. from what i remember the ko functions will always pass $data, event whether you specify it or not. to make it more explicit we could add , event into the function variables and where it is called.

src/RightClick.ts
```diff
@@ -705,6 +737,7 @@ export class RightClick {
             }
         }
         // adding a listener to function options that closes the menu if an option is clicked
-        $('#customContextMenu a').on('click',function(){if($(event.target).parents('.searchBarContainer').length){return}RightClick.closeCustomContextMenu(true)})
+        // TODO: get event from somewhere instead of global
+        $('#customContextMenu a').on('click',function(){if($((event as any).target).parents('.searchBarContainer').length){return}RightClick.closeCustomContextMenu(true)})
```
M-Wicenec: event as any? can we make this one more explicit

src/SideWindow.ts
```diff
@@ -184,37 +223,51 @@ export class SideWindow {
...
+        const leftWindowWidthSetting = Setting.find(Setting.LEFT_WINDOW_WIDTH);
+        const rightWindowWidthSetting = Setting.find(Setting.RIGHT_WINDOW_WIDTH);
+        const bottomWindowHeightSetting = Setting.find(Setting.BOTTOM_WINDOW_HEIGHT);
+
+        if (typeof leftWindowWidthSetting === "undefined" || typeof rightWindowWidthSetting === "undefined" || typeof bottomWindowHeightSetting === "undefined"){
+            console.warn("One or more window settings are undefined!");
+            return false;
+        }
+
+        const leftWindowWidth : number = leftWindowWidthSetting.value() as number;
```
M-Wicenec: do we need to cast these to number?

src/SideWindow.ts
```diff
@@ -184,37 +223,51 @@ export class SideWindow {
...
+        const rightWindowWidth : number = rightWindowWidthSetting.value() as number;
+        const bottomWindowHeight : number = bottomWindowHeightSetting.value() as number;
+
         if (isNaN(eagle.leftWindow().size())){
             console.warn("Had to reset left window width from invalid state (NaN)!");
-            eagle.leftWindow().size(Setting.find(Setting.LEFT_WINDOW_WIDTH).getPerpetualDefaultVal());
+            eagle.leftWindow().size(leftWindowWidth);
```
M-Wicenec, line 238: this seems different from before. before we where fetching the "perpetualdefaultvalue" when the width of a window is nan, to reset the value back to default.

src/Utils.ts
```diff
@@ -2793,7 +2872,9 @@ export class Utils {
     }
 
     static snapToGrid(coord: number, offset: number) : number {
-        const gridSize = Setting.findValue(Setting.SNAP_TO_GRID_SIZE);
+        const gridSizeSetting = Setting.find(Setting.SNAP_TO_GRID_SIZE);
+        const gridSize : number = gridSizeSetting ? gridSizeSetting.value() as number : 10;
```
M-Wicenec: as number may not be necessary? can we just use findValue? or do we need to check the setting exists first?

M-Wicenec (changes requested):
a few things worth taking another look at. loads of changes...  good work!

src/GraphRenderer.ts
james-strauss-uwa: Added a warning. But it does seem necessary to set x=0, otherwise the return value is null.

src/RightClick.ts
james-strauss-uwa: This function (edgeDropCreateNode) is not called from knockout, only from javascript. So there is no "event" parameter. Instead I get the event from window.event.

src/Node.ts
james-strauss-uwa: I checked it out. No, we don't need those two lines. The same thing is done by the line that is there now:

`node.setInputApplication(Node.fromOJSJson(nodeData.inputApplication, errorsWarnings, isPaletteNode));`

src/Node.ts
james-strauss-uwa: Same here.

src/Node.ts
james-strauss-uwa: It was not correct.

But reverting back to using Utils.generateNodeId() caused a circular import dependency that prevented EAGLE from loading. Node depended on Utils, which depended on Node.

To fix this I moved all the Id generation code into a new file (src/Id.ts).

james-strauss-uwa:
@M-Wicenec this is ready for a review of the requested changes.

It took a 3 attempts to get the CI tests to succeed, but otherwise good.

PR #1002, opened 2026-02-25
EAGLE-1556: Script to update old graphs

james-strauss-uwa:
A new feature that uses EAGLE to load old graphs and save them in the latest format. Effectively updating old graphs.

This initially called for a playwright script, but it seemed useful to allow the user to select the source and destination repositories/branches without having to edit the script each time.

The basics of this are working, I just need some feedback on the remaining parts:
- Is the current menu location good. It is in a new Help > Tools menu. Is there a better place for it? Will we ever have other tools here?
- Do we need more feedback on success rate? For example, if one or more graphs fails to update, it just displays "Error" in the table row(s). Do we need to make that more obvious?
- At the moment, the user must have already created a new destination branch in their destination repository. The user must also have added this repository/branch to the list of repos in EAGLE. Creating the repo/branch must be done outside EAGLE. Is this clear to the user? Should we add code to allow the user to specify a new branch and create it at runtime?
- Any feedback on the UI/UX flow.

## Summary by Sourcery

Add a UI-driven tool for updating graphs from old repositories to the latest format and batch-pushing them to remote Git repositories.

New Features:
- Introduce a Graph Updater modal that lets users select source and destination repositories, fetch all graphs, update them to the latest schema, and push changes in bulk.
- Add keyboard shortcut and Help > Tools menu entry to launch the Graph Updater workflow from the UI.
- Support batch-saving multiple graph files to remote GitHub/GitLab repositories via new backend endpoints and a client helper.

Enhancements:
- Extend repository browsing utilities to expand all folders and collect all graph files, and expose full path names for repository files.
- Improve GitHub save error handling by catching metadata-setting failures and returning structured errors.

## Summary by Sourcery

Introduce a UI-driven Graph Updater workflow to batch-convert legacy graph files to the latest schema and push them to selected Git repositories.

New Features:
- Add a Graph Updater modal that lets users choose source and destination repositories, fetch all graph files, convert them to the latest format, and push them in bulk.
- Expose the Graph Updater via a Help > Tools menu entry and a keyboard shortcut.
- Implement backend endpoints and client support for saving multiple graph files to remote GitHub/GitLab repositories in a single commit.

Enhancements:
- Extend repository browsing to expand all folders and collect all graph files, and expose full path names for repository files.
- Improve GitHub save error handling by catching metadata-setting failures and returning structured errors for UI display.


SourceryAI (commented):
Hey - I've found 3 issues, and left some high level feedback:

- In `GraphUpdater.push`, the function is declared as returning `Promise<void>` but in the error path it `return`s `errorJSON.error`; this is inconsistent with the signature and will either cause a type error or silently drop the value—either change the return type or stop returning a value from this method.
- The error handling in `GraphUpdater.push` assumes `error` is a JSON string (`JSON.parse(error)`), but `saveFilesToRemote` rejects with an error object; this will itself throw and mask the original failure, so consider standardizing the error shape (e.g. rejecting with a structured object/string) and avoiding `JSON.parse` on non-strings.
- In `GraphUpdater.fetchLogicalGraphs` and `GraphUpdater.push`, the selected repository index is taken via `parseInt(...val() as string)` with no validation; it would be safer to guard against `NaN`/out-of-range indices before indexing into `Repositories.repositories()` so the UI can show a clear error instead of potentially throwing.

<details>
<summary>Prompt for AI Agents</summary>

~~~markdown
Please address the comments from this code review:

## Overall Comments
- In `GraphUpdater.push`, the function is declared as returning `Promise<void>` but in the error path it `return`s `errorJSON.error`; this is inconsistent with the signature and will either cause a type error or silently drop the value—either change the return type or stop returning a value from this method.
- The error handling in `GraphUpdater.push` assumes `error` is a JSON string (`JSON.parse(error)`), but `saveFilesToRemote` rejects with an error object; this will itself throw and mask the original failure, so consider standardizing the error shape (e.g. rejecting with a structured object/string) and avoiding `JSON.parse` on non-strings.
- In `GraphUpdater.fetchLogicalGraphs` and `GraphUpdater.push`, the selected repository index is taken via `parseInt(...val() as string)` with no validation; it would be safer to guard against `NaN`/out-of-range indices before indexing into `Repositories.repositories()` so the UI can show a clear error instead of potentially throwing.

## Individual Comments

### Comment 1
<location path="src/GraphUpdater.ts" line_range="229" />
<code_context>
+        this.setState(true, false, false, false);
+
+        // get source repository
+        const srcRepoIndex = parseInt($('#graphUpdaterModalSourceRepositorySelect').val() as string);
+        const srcRepo = Repositories.repositories()[srcRepoIndex];
+        if (srcRepo === null){
</code_context>
<issue_to_address>
**issue:** Handle invalid/empty repository selection more defensively.

If no option is selected, `$('#graphUpdaterModalSourceRepositorySelect').val()` can be `null`, so `parseInt` returns `NaN` and `Repositories.repositories()[NaN]` is `undefined`. The `if (srcRepo === null)` check won’t catch this. Please guard for `null`/`undefined` and `Number.isNaN(srcRepoIndex)` before indexing into `Repositories.repositories()`. The same check should be applied to the destination repo in `push()`, which also only checks for `null`.
</issue_to_address>

### Comment 2
<location path="src/GraphUpdater.ts" line_range="280" />
<code_context>
+            graphFile.state(GraphUpdater.FileStatus.Updating);
+
+            // fetch the file data
+            const fileData: string = await openRemoteFileFunc(graphFile.file().repository.service, graphFile.file().repository.name, graphFile.file().repository.branch, graphFile.file().path, graphFile.file().name);
+
+            // determine if graph is OJS or V4
</code_context>
<issue_to_address>
**issue (bug_risk):** Wrap remote file load and parsing in a try/catch to avoid leaving the UI in a partially-updating state.

If `openRemoteFileFunc` rejects or `JSON.parse(fileData)` throws, the error will escape `update()` and the `GraphUpdaterFile` never reaches an `Error` state, so the modal spinner may never clear. Please wrap the remote load + parse + conversion for each file in try/catch, set `graphFile.state(FileStatus.Error)` on failure, and continue with remaining files. An outer try/finally to keep the global `setState` flags consistent even when a file fails would also help.
</issue_to_address>

### Comment 3
<location path="src/GraphUpdater.ts" line_range="383-388" />
<code_context>
+        } catch (error) {
</code_context>
<issue_to_address>
**issue (bug_risk):** Avoid assuming a JSON-encoded error and returning a string from a `Promise<void>` function.

Inside `push()`, the `catch` assumes `error` is JSON text (`JSON.parse(error)`), but `Eagle.saveFilesToRemote` rejects with the raw error object, so this will likely throw inside the handler. It also returns `errorJSON.error` (a string) from a `Promise<void>`, which is inconsistent with the signature. Instead, handle `error` as an unknown object (e.g., log / stringify it) and either rethrow or return `void` after showing the message—no return value is needed here.
</issue_to_address>
~~~

</details>

***

Hi @james-strauss-uwa! 👋

Thanks for trying out Sourcery by commenting with `@sourcery-ai review`! 🚀

Install the [sourcery-ai bot](https://github.com/apps/sourcery-ai/installations/new) to get automatic code reviews on every pull request ✨

<sub>
Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.
</sub>

src/GraphUpdater.ts
```diff
@@ -158,132 +179,224 @@ export class GraphUpdater {
...
+        this.updatedLogicalGraphs.removeAll();
+        this.setState(false, false, false, false);
+    }
+
+    static async fetchLogicalGraphs(): Promise<void> {
+        console.log("GraphUpdater.fetchLogicalGraphs()");
+        this.setState(true, false, false, false);
+
+        // get source repository
+        const srcRepoIndex = parseInt($('#graphUpdaterModalSourceRepositorySelect').val() as string);
```
SourceryAI, line 272: **issue:** Handle invalid/empty repository selection more defensively.

If no option is selected, `$('#graphUpdaterModalSourceRepositorySelect').val()` can be `null`, so `parseInt` returns `NaN` and `Repositories.repositories()[NaN]` is `undefined`. The `if (srcRepo === null)` check won’t catch this. Please guard for `null`/`undefined` and `Number.isNaN(srcRepoIndex)` before indexing into `Repositories.repositories()`. The same check should be applied to the destination repo in `push()`, which also only checks for `null`.

✅ Addressed in a4aaa9051f7bc46d37034fb0973374ff6e005dcc: The source repository selection logic has been updated to guard against `NaN` indices and null/undefined repositories, addressing invalid/empty selection cases.

james-strauss-uwa:
Changes after talking to Andreas/Moritz:

- Better description of what the Tool is going to do. At the moment it appears as though it might be making changes in-place!
- Give the user more information than just "Success". Perhaps number of errors/warnings pre and post update. Possibly also check that number of nodes/edges etc has not changed.
- Add a checkbox to specify whether to apply fixes to the updated graphs.
- Allow users to view the old and new graphs from the interface. Open graphs in new tab, in a "read-only" UI mode?
- Add a per-graph checkbox to allow users to specify which graphs will be updated. Default to all.
- Add a per-graph checkbox to allow users to specify which graphs will be committed to the destination branch. Default to all successful graphs.
- Handle timeouts due to GitHub rate limiting.
- Move menu item to top of section

james-strauss-uwa:
@M-Wicenec this is ready for another review

I've implemented the stuff we spoke about last time we reviewed:
- More information about what the tool does, in a collapsable section at the top.
- Leave modal open after push, to give the user feedback, and to allow the user to open the updated graphs (in a new tab).
- Added option to automatically create a new branch as the destination for updated graphs
- Added columns to the "results" table with more info on the graphs, errors pre/post fix etc
- Better handling and display of any errors that occurred during updating

The thing we talked about that I haven't done:
- Add the option to automatically merge the changes from EAGLE. This might be a phase 2. It actually requires first creating a pull request, then waiting for any CI to pass/fail, then trigger the merge. Plus it would have to support github and gitlab. So I think it is too much additional work.

M-Wicenec:
fix graphs checkbox should be enabled by default

after updating the graphs the number of found errors and warnings does not match the number of errors and warnings when you open said graph

i then told it to push to a seperate repo with fixes enabled, it worked, the repo showed it was updated  but the errors in the graphs on the repo were not fixed. this may be related to the previous point

one graph no longer opens in eagle after updating the repo. no errors are shown in the console or eagleserver windwow. graph: (http://localhost:8888/?service=GitHub&repository=M-Wicenec/EagleTest&branch=main&path=&filename=chilies_daliuge_split_organised.graph)

the show updated graph column does not need to show the full file name again, it can be a simple 'load' button. hopefully this means we wont get a horizontal scroll bar

minor css:
scroll should use the 'light mode' version (we have a scroll bar for dark backgrounds eg in left window, and a scrollbar for light backgrounds used in right window)
table header should be floating
the table should not be a set height. currently when there is not enough space we get an odd double scroll bar scenario. and when there is more space, the table doesnt use all of the space available to it.

potential future upgrade:
user should able to open the updated graphs in another tab from memory before pushing. can this be done similarly to the copy paste across tabs feature?

PR #1014, opened 2026-03-26
EAGLE-1630: Pin Typescript version to 5.8.x

james-strauss-uwa:
Typescript 6.0 has been released, and compiling EAGLE under that version raises warnings about features deprecated in version 7.0.

We'll have to update EAGLE to be ready for Typescript 7.0 at some stage, but for the moment, just stay with 5.8.x.

I've added one more commit since the initial work. The Typescript compile started to fail again. It seems the Typescript compiler was version 6.0.2 even though we pin the version to 5.8.3. I fixed this by running the compiler using "npx tsc" instead of "tsc". This makes sure the version of tsc installed by npm (5.8.3) is run, instead of some global version (6.0.2) that is somehow the default on the system.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -26,7 +26,9 @@ jobs:
       - name: Install python setuptools
         run: pip install setuptools==81.0.0
       - name: Compile Typescript
-        run: tsc
+        run: |
+          npx tsc --version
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
Looks all good!

PR #1015, opened 2026-04-01
EAGLE-1636: Improved Node.getPorts()

james-strauss-uwa:
Just simplified the code to remove the duplicate ports in the results.

@M-Wicenec just wanted you to check that the previous behaviour wasn't required for the renderer (or anything else). Thanks.

## Summary by Sourcery

Enhancements:
- Refactor Node.getPorts() to iterate over all fields and return only those with a port usage, eliminating previous duplicate-handling logic.


M-Wicenec (approved):
I think its good. The relying features are working as expected

PR #1016, opened 2026-04-01
EAGLE-1635: Fixed bug when user specifies a commit message

james-strauss-uwa:
Actually, it seems that this doesn't always cause a bug, but the function call should use await anyway.

## Summary by Sourcery

Bug Fixes:
- Fix potential race condition or unhandled behavior when requesting a user-entered commit message by properly awaiting the asynchronous prompt call.


james-strauss-uwa:
@sourcery-ai review

M-Wicenec (approved):
looks good, seems to be working for me.

PR #1020, opened 2026-04-09
TS6 compatibility

M-Wicenec:
the latest builds of Visual Studio Code is using TS6. This has caused many errors being highlighted in various eagle files.

changes to various functions to comply with strict types, which are enabled by default on TS6
fixes to import issues cause by strict checks in TS6
addition of tsconfig.json file in e2e (the playwright test folder)
`@ts-nocheck `added in require-config.ts to prevent an error at runtime, complaining about require being declared twice
`"skipLibCheck": true `added to prevent ts6 from compiling packages we use such as knockout. this was causing many errors at runtime.

## Summary by Sourcery

Update TypeScript configuration and typings across the app and e2e tests to be compatible with TS6 and stricter type checking.

Enhancements:
- Tighten typings in e2e TestHelpers, including explicit Playwright Page usage, Node HTTP/FS types, and safer object utilities.
- Improve runtime error handling in Eagle by normalizing caught errors to strings/messages before displaying them to users.
- Use concrete NodeJS.Timeout types for tutorial timers to align with Node type definitions.

Build:
- Adjust main tsconfig to set rootDir, enable skipLibCheck, add Node types, and ignore TS 6.0 deprecations for smoother compilation.
- Move TypeScript to devDependencies and bump to a TS 5.9.x version compatible with TS6 tooling.
- Add a dedicated tsconfig for the e2e Playwright folder with appropriate libs, module resolution, and type roots.

Chores:
- Disable TypeScript checking in require-config to avoid duplicate require declaration errors under stricter TS settings.


james-strauss-uwa (approved):
Looks good, thanks.

PR #1022, opened 2026-04-13
added correct and updated imports for jqueryUI and jqueryMigrate and …

M-Wicenec:
…their types

## Summary by Sourcery

Align jQuery-related module names and type definitions across the project and clean up redundant modal draggable setup.

Bug Fixes:
- Correct jQuery Migrate RequireJS path alias to match the actual script name.

Enhancements:
- Add TypeScript type definitions for jQuery UI and jQuery Migrate and register them in the compiler configuration.
- Remove a duplicated modal draggable initialization handler from the main entry script while retaining it in the central Eagle initialization.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,8 @@
         "@types/bootstrap": "^5.1.4",
         "@types/bootstrap-notify": "^3.1.34",
         "@types/jquery": "^3.3.38",
+        "@types/jquery-migrate": "^3.3.3",
+        "@types/jqueryui": "^1.12.24",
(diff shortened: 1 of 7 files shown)


james-strauss-uwa (approved):
Looks good. I just made a change to avoid one "any" cast.

Can we also make a JIRA ticket to investigate the Eagle.ts code, where a handler is added for "shown.bs.modal". I'd like to try calling draggable once for each modal, instead of on every modal, every time a modal is shown. Low priority I guess.

PR #1025, opened 2026-04-16
EAGLE-1642: LogicalGraph.clone() cloned graph edges reference original graph nodes and ports

james-strauss-uwa:
Fixed a bug in LogicalGraph.clone() where the cloned graph edges referenced nodes and ports from the original graph.

Also added a unit test to make sure the fix to LogicalGraph.clone() works. Unfortunately the unit tests use the "any" type in several places. This is consistent with several of our other playwright tests, but isn't ideal. I investigated a few approaches to using real Eagle types (Node, Edge, Field etc) within the unit test, but nothing I tried worked.

## Summary by Sourcery

Fix LogicalGraph cloning so that cloned edges reference the cloned graph’s nodes and ports instead of the originals and add a regression test to cover the behavior.

Bug Fixes:
- Ensure edges in a cloned LogicalGraph are rewired to reference nodes and ports from the cloned graph rather than the original graph.

Tests:
- Add an end-to-end test verifying that LogicalGraph.clone() produces edges correctly wired to cloned nodes and ports.


M-Wicenec (approved):
looks good, new test is reliable

james-strauss-uwa:
New test failed once. Worked when re-tried. The failure was in an unrelated part of the code, the attempt to drag an edge between the HelloWorldApp and File nodes.

PR #1027, opened 2026-04-22
EAGLE-1643: Fix duplicate snapshot detection in Undo system

james-strauss-uwa:
Previously it was checking to see whether two newly created objects were identical, which could never be true.

Now we store a hash along with each Undo snapshot, and can check whether a newly created snapshot matches the snapshot already stored at the top of the Undo system.

## Summary by Sourcery

Improve Undo snapshot handling to avoid storing duplicate snapshots for unchanged graph content.

Bug Fixes:
- Fix Undo system duplicate detection so snapshots are considered identical based on a stored hash of their serialized content rather than object identity.

Tests:
- Add an end-to-end test that verifies the Undo system does not create duplicate snapshots when content has not changed.

### DIFF
--- a/e2e/undoDuplicateDetection.spec.ts
+++ b/e2e/undoDuplicateDetection.spec.ts
@@ -1,0 +1,53 @@
+import { test, expect } from '@playwright/test';
+import { TestHelpers } from './TestHelpers';
+
+test('Undo duplicate snapshot detection', async ({ page }) => {
+    await page.goto('http://localhost:8888/?tutorial=none');
+    await expect(page).toHaveTitle(/EAGLE/);
+
+    // set 'Expert' UI mode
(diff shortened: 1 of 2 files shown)


james-strauss-uwa:
Apparently:

> In modern JavaScript, JSON.stringify() follows a well-defined, predictable key ordering based on the same algorithm as Object.keys(). While the [official JSON specification](https://www.json.org/json-en.html) describes an object as an "unordered collection," JavaScript engines (like V8 in Chrome/Node.js) enforce specific rules for property traversal during serialization.

So the key order will be deterministic.

The 32-bit hash has a ~1-in-4-billion chance of a false positive collision, which would silently drop a real state change from the undo stack. So unlikely, and with small consequences.

M-Wicenec (approved):
looks good, test is dependable

PR #1028, opened 2026-04-23
EAGLE-1644: Sanitize HTML before attaching to DOM

james-strauss-uwa:
Multiple locations could be injected with untrusted content as raw HTML via jQuery .html()

Switched to using .text() where possible.

The remaining locations using .html() now sanitize the HTML before attaching to the DOM.

The sanitise function:

1. Creates a element hierarchy from the HTML string
2. Walks the hierarchy removing unknown elements
3. Creates a HTML string from the sanitized hierarchy

The sanitize function uses a white-list of allowed html tags/attributes, so may have to be updated in future. If we find some documentation that uses html tags not found in our white-list, the information could be removed! Actually, this isn't too bad, I've made changes to retain sanitized html from within dangerous tags where possible.

Also removed two unused comment/description related functions `getDescriptionHTML()` and `getCommentHTML()`

## Summary by Sourcery

Sanitize HTML content before inserting it into the DOM and tighten handling of user-facing messages and markdown rendering.

New Features:
- Add a reusable HTML sanitization helper that whitelists allowed tags and attributes and enforces safe URL schemes before rendering HTML.

Bug Fixes:
- Prevent injection of untrusted HTML in user message and markdown rendering flows by using plain text rendering by default and sanitizing when HTML is required.

Enhancements:
- Update user message utilities and markdown conversion to integrate the HTML sanitizer and add an option for explicitly rendering sanitized HTML.
- Remove unused description and comment HTML helpers from the Node model.


M-Wicenec (approved):
looks good, prevents tooltip pop ups from running eg. script tags. 

ive uncovered a related issue with script tags in node or field names. Up to you if you want this to be a separate ticket.

when adding a `<script>console.log('hi')</script>` into the node name, it will run the script when eg. right clicking the node in the canvas or duplicating the node. When adding this script to a field attribute name, it triggers the script when favoriting the field, and using the edit button on that field in the inspector

james-strauss-uwa:
I've made a change to solve the right-click issue mentioned above. There was one more place where the node name and description were added directly to the html (the right click menu). I've wrapped these places in calls to Utils.markdown2html(), which calls Utils.sanitizeHtml().

james-strauss-uwa:
@M-Wicenec I got the AI to review the code and find all the places where raw user-entered values were added to the HTML (without sanitization). It found the field attribute name issue you described above. So you can no longer run script code by favoriting the field.

It didn't catch the issue you had when clicking the "edit field value" button. I had to fix that separately.

PR #1029, opened 2026-05-06
EAGLE-1628: Remove code using pkg_resources and remove dependency of setuptools version 81.

james-strauss-uwa:
> The pkg_resources module, formerly part of the setuptools library, has been officially removed as of setuptools version 82.0.0 (released around February 2026)

This branch removes code using the pkg_resources module, and allows the unit tests to use the most recent version of setuptools, instead of pinning the version to 81.0.0.

Please make sure that pkg_resources is not required for the deployment on AWS.

## Summary by Sourcery

Remove reliance on pkg_resources for locating server templates/static/src directories and allow CI Playwright workflow to use the latest setuptools instead of pinning to version 81.

Enhancements:
- Replace pkg_resources-based resource path resolution with filesystem-based paths derived from the eagleServer module location.

CI:
- Update Playwright GitHub Actions workflow to install the latest setuptools rather than a fixed 81.0.0 version.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -24,7 +24,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install .
       - name: Install python setuptools
-        run: pip install setuptools==81.0.0
+        run: pip install setuptools
       - name: Compile Typescript
         run: |
           npx tsc --version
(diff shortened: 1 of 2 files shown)


M-Wicenec (approved):
looks good from my side

awicenec (approved):
I've checked that the deployment on AWS is not affected by this change. Note that the AWS deployment does not even install or run python directly, but just the slimmed docker image.

PR #1030, opened 2026-05-07
EAGLE-1648: Allow LogicalGraphs to be checked from memory when not loaded as the Eagle active graph

james-strauss-uwa:
Preparatory work for EAGLE-1556 (Update old graphs script) which needs to test graphs after updating, to notify the user of errors in the resulting graph.

Renaming checkGraph functions/ids to make their purpose clearer, since it is actually more like 'checkEagle'.

Refactor of LogicalGraph.isValid to make it possible to check graphs even when they are not loaded into Eagle as the currently edited graph.

## Summary by Sourcery

Refactor graph validation to support checking logical graphs independently of the active Eagle instance and rename related APIs and UI bindings from graph-specific to Eagle-wide terminology.

Enhancements:
- Rename graph validation APIs, keyboard shortcut IDs, UI bindings, and CSS hooks from graph-focused (checkGraph/GraphErrors) to Eagle-focused (checkEagle/EagleErrors) terminology for clarity.
- Update LogicalGraph validity checks to accept an explicit graph and optional Eagle instance so graphs can be validated even when not loaded as the active graph.
- Adjust error display flows, inspectors, tutorials, and modals to use the new Eagle-wide validation entry points and bottom-window mode.

### DIFF
--- a/e2e/TestHelpers.ts
+++ b/e2e/TestHelpers.ts
@@ -39,14 +39,14 @@ export class TestHelpers {
     static async setShortDescription(page: Page, description: string): Promise<void> {
         await page.evaluate( (description: string) => {
             (window as any).eagle.logicalGraph().fileInfo().shortDescription = description;
-            (window as any).eagle.checkGraph();
+            (window as any).eagle.checkEagle();
         }, description);
     }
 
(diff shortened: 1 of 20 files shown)


M-Wicenec (approved):
looks all good, still working properly

PR #1032, opened 2026-05-26
EAGLE-1649: Improve reliability of cloneLogicalGraph.spec.ts unit test

james-strauss-uwa:
At the moment, this test is failing more often than it succeeds in Github CI.

This AI suggested change finds the boundingBox for each port, and makes sure to drag from and to the middle of the bounding boxes. It seems to be more reliable.

It also adds a clear Error if it could not locate either the source or destination ports.

## Summary by Sourcery

Improve reliability of the LogicalGraph.clone end-to-end test when creating edges between nodes.

Bug Fixes:
- Prevent flaky failures in the cloneLogicalGraph end-to-end test by making the drag operation more robust and explicit when connecting ports.

Tests:
- Update the cloneLogicalGraph end-to-end test to drag from the center of computed port bounding boxes and surface a clear error when ports cannot be located.

### DIFF
--- a/e2e/cloneLogicalGraph.spec.ts
+++ b/e2e/cloneLogicalGraph.spec.ts
@@ -24,7 +24,6 @@ test('LogicalGraph.clone() does not share references with original', async ({ pa
 
     // center the graph
     await page.getByRole('button', { name: 'filter_center_focus' }).click();
-    await page.waitForTimeout(200);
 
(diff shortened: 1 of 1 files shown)


M-Wicenec (approved):
nice, ive run it a bunch of times and it seems reliable. ive created a ticket for the edge drawing helper function, so we can get this fix live

PR #1033, opened 2026-05-26
EAGLE-1653: Readthedocs publishing cancelled unless commit has a tag

james-strauss-uwa:
Only publish to readthedocs if the commit is tagged. This will prevent issues like we have at the moment, where the features in master, but not released, are already documented on readthedocs.

## Summary by Sourcery

Build:
- Add a post-checkout job in the Read the Docs configuration that cancels documentation builds when the current commit is not an exact Git tag.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,10 +9,14 @@ build:
     python: "3.10"
   jobs:
     post_checkout:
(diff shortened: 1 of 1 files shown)


james-strauss-uwa:
I think the plan is to use some code from DALiuGE instead of a custom solution like the one here

PR #1037, opened 2026-06-15
EAGLE-1660: Update showdown dependency

james-strauss-uwa:
Dependabot flagged that showdown (the markdown-to-html renderer) had an outstanding security issue. It turns out that showdown could not be updated to the latest version, since it has not been actively developed, and is considered abandoned.

The most popular alternative is a package called marked.

So I replaced showdown with marked, then saw there was an official extension for marked called marked-highlight that does syntax highlighting of code blocks. I added that too. The syntax marked-highlight also depends on an external highlighting package, and the most common in highlight.js, so that is also included.

There was a bit of work required to import highlight.js correctly. I ended up using the CDN-ready version of highlight.js, and defining a "shim" in the require.js config to load it correctly.

## Summary by Sourcery

Replace the deprecated showdown markdown renderer with the marked ecosystem and wire it into the app for markdown rendering and code highlighting.

Enhancements:
- Switch markdown rendering from showdown to marked and integrate marked-highlight for syntax highlighting in the markdown-to-HTML utility.
- Hook highlight.js into the AMD/RequireJS configuration so it can be used as the highlighter backend for marked-highlight.
- Add GitHub-style syntax highlighting styles and bundle the marked, marked-highlight, and highlight.js browser builds into the static assets.

Build:
- Update package.json and lockfile to remove showdown and its types and add marked, marked-highlight, highlight.js, and @highlightjs/cdn-assets as dependencies.


M-Wicenec (approved):
found no issues, seems to work as intended

PR #1039, opened 2026-06-16
EAGLE-1645: Fix JSON injection issue

james-strauss-uwa:
AI suggested that createCommitJsonString could be used to inject arbitrary data:

The function builds JSON by concatenating repository.branch, token, commitMessage, etc. into strings without escaping. A commit message containing " breaks the JSON structure and can inject arbitrary fields.

Now we stringify all the inputs (branch, token, commitMessage etc) so that the concatenation is done correctly.

## Summary by Sourcery

Bug Fixes:
- Prevent JSON injection and malformed payloads in createCommitJsonString by properly escaping commit metadata fields before concatenation.


M-Wicenec (approved):
nice, thats a good change

PR #1040, opened 2026-06-17
EAGLE-1631: Anonymous file loading from public git repos if credentials are missing/incorrect

james-strauss-uwa:
If the user has not set their github/gitlab credentials, and tries to load a file from a git repo, then EAGLE used to fail.

Now the eagleServer will fall back to attempting to fetch the file anonymously, which should work fine for public repositories.

You can test this by temporarily removing your github/gitlab keys (or making them incorrect). NOTE: you'll pretty quickly hit the github API limit for anonymous access.

## Summary by Sourcery

Allow anonymous access to public GitHub and GitLab repositories when configured credentials are missing or invalid, and surface this behavior to the user interface.

New Features:
- Fallback to anonymous access for public GitHub repositories when an access token is missing or invalid while listing files or opening a file.
- Fallback to anonymous access for public GitLab repositories when an access token is missing or invalid while listing files or opening a file.

Enhancements:
- Include a credentialsIgnored flag in GitHub and GitLab API responses so the UI can indicate when invalid credentials were bypassed.
- Update the GitHub and GitLab front-end clients to handle the new response shape and display warning notifications when tokens are ignored.
- Clarify file-load warning notifications to mention the originating service (GitHub or GitLab) in addition to the filename.

### DIFF
--- a/eagleServer/eagleServer.py
+++ b/eagleServer/eagleServer.py
@@ -302,6 +302,7 @@ def get_git_hub_files_all():
     folder_name, repo_name = extract_folder_and_repo_names(repo_name)
 
     # authenticate or not
+    credentials_ignored = False
     g = github.Github(repo_token) if repo_token else github.Github()
 
     try:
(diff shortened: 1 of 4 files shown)


M-Wicenec (approved):
yep, this is working. bad credentials works as well. code looks good

PR #1044, opened 2026-06-25
EAGLE-1639: Discourage users for saving files in old OJS format

james-strauss-uwa:
Before saving to git or disk, we now check the Setting.DALIUGE_SCHEMA_VERSION setting. If the setting is not V4, we popup a confirmation dialog checking that the user actually wants to save using the older format.

Users can click the "Don't ask me again" checkbox to avoid being asked repeatedly if they want to continue using the old format. (Setting.CONFIRM_OJS_FORMAT).

## Summary by Sourcery

Add user confirmation when saving files using the legacy OJS schema format and provide a setting to control this warning.

New Features:
- Prompt users before saving graphs and palettes in the older OJS format when a newer V4 format is available.
- Introduce a configurable setting to suppress or re-enable warnings about saving in the OJS format.

Enhancements:
- Reset the new OJS format confirmation setting alongside other confirmation pop-up settings.

### DIFF
--- a/src/Setting.ts
+++ b/src/Setting.ts
@@ -352,6 +352,7 @@ export class Setting {
     static readonly CONFIRM_REMOVE_REPOSITORIES : string = "ConfirmRemoveRepositories";
     static readonly CONFIRM_DELETE_FILES : string = "ConfirmDeleteFiles";
     static readonly CONFIRM_DELETE_OBJECTS : string = "ConfirmDeleteObjects";
+    static readonly CONFIRM_OJS_FORMAT : string = "ConfirmOjsFormat";
     static readonly TEST_TRANSLATE_MODE : string = "TestTranslateMode";
 
     static readonly SHOW_DEVELOPER_NOTIFICATIONS: string = "ShowDeveloperNotifications";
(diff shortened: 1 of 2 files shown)


src/Setting.ts
```diff
@@ -451,6 +452,7 @@ const settings : SettingsGroup[] = [
             new Setting(false, "Confirm Remove Repositories", Setting.CONFIRM_REMOVE_REPOSITORIES, "Prompt user to confirm removing a repository from the list of known repositories.",false , Setting.Type.Boolean, true,true,true,true,true),
             new Setting(false, "Confirm Delete Files", Setting.CONFIRM_DELETE_FILES, "Prompt user to confirm when deleting files from a repository.", false, Setting.Type.Boolean, true,true,true,true,true),
             new Setting(false, "Confirm Delete Objects", Setting.CONFIRM_DELETE_OBJECTS, "Prompt user to confirm when deleting node(s) or edge(s) from a graph.",false , Setting.Type.Boolean, true,true,true,true,true),
+            new Setting(false, "Warn On OJS Format Save", Setting.CONFIRM_OJS_FORMAT, "Warn user when saving in the older OJS format that the newer V4 format is recommended.", false, Setting.Type.Boolean, true,true,true,true,true),
```
M-Wicenec, line 455: I've just realised all of these confirm states could be set to perpetual. This just means that they arent saved individually for each ui mode.

M-Wicenec (approved):
Looks good, one minor comment, probably dont have to do it as part of this ticket

src/Setting.ts
james-strauss-uwa: @M-Wicenec OK, I'll add a new JIRA ticket for this (EAGLE-1669)

PR #1046, opened 2026-06-29
EAGLE-1667: Fixed undo bug where full state of graph was not written to snapshots

james-strauss-uwa:
Added a new playwright test to check the specific scenario mentioned in the bug report.

Fixed serialisation of Field and FileInfo
- Field now includes "changeable" boolean in serialisation. "changeable" was always intended to be a run-time attribute only, but if the attribute is not serialised, then the value is lost when written to an Undo snapshot. Therefore "Redo" will not have the correct value, and will introduce a bunch of errors/warnings into the restored graph.
- FileInfo now serialises LogicalGraphs loaded from Urls.

## Summary by Sourcery

Ensure undo snapshots fully capture graph state after fixing errors and loading graphs from URLs.

Bug Fixes:
- Prevent re-entrant or rapid consecutive fixAll invocations from creating multiple partial undo snapshots.
- Ensure fixAll pushes an undo snapshot only when fixes are actually applied, maintaining correct undo boundaries.
- Include the Field.changeable flag in field serialization and deserialization so its state is preserved across undo/redo.
- Persist repository service and URL-backed LogicalGraph location consistently in FileInfo so URL-loaded graphs survive undo/redo without location changes.

Tests:
- Add a Playwright regression test verifying that undo after fixAll does not reintroduce previously fixed graph errors or warnings.


M-Wicenec (approved):
works well. 
thats a cool test! i assume you prompted ai to create that first so you can then let it attempt to fix the issue while being able to test itself?

james-strauss-uwa:
> works well. thats a cool test! i assume you prompted ai to create that first so you can then let it attempt to fix the issue while being able to test itself?

@M-Wicenec Yes, you're right. I think I pretty much copypasted your description of how to reproduce the issue into AI and told it to produce a playwright test. It found the fixes too.

PR #1048, opened 2026-07-01
EAGLE-1669: Make the 'confirm' settings perpetual

james-strauss-uwa:
Also modified the wording of the OJS/V4 confirm setting for consistency

## Summary by Sourcery

Make action confirmation-related settings persistent and align wording for the OJS/V4 save confirmation option.

Enhancements:
- Enable the reset action confirmations button and all confirmation-related settings to be treated as persistent options rather than transient prompts.
- Update the description text for the OJS format save confirmation to clearly state that it warns when saving in the older format and recommends the V4 format.


M-Wicenec (approved):
👍

PR #1049, opened 2026-07-01
Eagle 1662

M-Wicenec:
added a test that runs through all of the tutorials. only using the interface (it is not allowed to take shortcuts by prompting eagle directly)

the test uses the tutorial itself to figure out what to click/ do. which does mean that some more complex tutorial steps require a custom test function attached to them. This is something to keep in mind when writing tutorials.

## Summary by Sourcery

Add an end-to-end Playwright test that exercises all registered tutorials via the UI and extend the tutorial system with test hooks to make complex steps automatable.

New Features:
- Introduce a Playwright e2e test that discovers all tutorials and runs each one through the application interface.
- Expose helper APIs on the tutorial system to list tutorials and their names for consumption by tests.

Enhancements:
- Add test-specific functions and metadata to tutorial steps so complex interactions (context menu, node selection, edge creation) can be driven reliably from automated tests.
- Refine tutorial configuration to attach test step hooks to several graph-related tutorials for better testability.
- Adjust Playwright configuration to use a consistent 1920x1080 viewport and scale factor across browsers.
- Update main app startup logic so an explicitly requested tutorial suppresses the Whats New modal to avoid overlay conflicts.
- Relax the comment field’s disabled binding in the config parameter table to better support automated input during tutorials.

Build:
- Standardize browser launch and viewport settings in Playwright configuration to improve reliability and determinism of UI tests.

Tests:
- Create comprehensive TestHelpers utilities for driving tutorial flows via Playwright, including running steps by type and interacting with graph nodes through the UI.
- Add a Tutorials e2e spec that boots the app, enumerates available tutorials, and verifies each one completes successfully without direct backend shortcuts.


e2e/TestHelpers.ts
```diff
@@ -2,10 +2,18 @@ import fs from 'fs';
 import https from 'https';
 import http from 'http';
 import path from 'path';
-import type { Page } from '@playwright/test';
-import { test, expect } from '@playwright/test';
+import { test, expect, type Page } from '@playwright/test';
 
 export class TestHelpers {
+    private static contextMenuAnchorIndex = 0;
```
james-strauss-uwa, line 14: Could you add some comments describing what **contentMenuAnchorIndex** and **lastContextMenuPoint** store, and why they are required.

e2e/TestHelpers.ts
```diff
@@ -2,10 +2,18 @@ import fs from 'fs';
...
+    private static lastContextMenuPoint: { x: number; y: number } | null = null;
+    private static readonly TutorialStepType = {
```
james-strauss-uwa, line 18: This looks the same as the TutorialStep.Type enum. Is there a reason we need to replicate it here?

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+
+    static async runTutorialPressStep(page: Page, title: string, selector: string): Promise<void> {
+        await test.step(`Tutorial press step: ${title}`, async () => {
+            await TestHelpers.waitForTutorialStep(page, title);
+            await page.locator(selector).click();
+            await page.locator('#tutorialInfoPopUp').waitFor({ state: 'detached', timeout: 5000 });
+        });
+    }
+
+    static async runTutorialByName(page: Page, tutorialName: string, maxSteps = 100): Promise<void> {
```
james-strauss-uwa: **maxSteps** here seems like a bit of a land-mine, waiting to explode if our tutorials ever exceed 100 steps. All-of-a-sudden our tutorial will fail to complete.

Is there a reason to abort after 100 steps?

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+        //.step is creating a test step. this is so we know exactly where we failed if something goes wrong.
+        await test.step(`Start tutorial: ${tutorialName}`, async () => {
+            await page.evaluate((name: string) => (window as any).TutorialSystem.initiateTutorial(name), tutorialName);
+            await page.locator('#tutorialInfoPopUp').waitFor({ state: 'attached', timeout: 10000 });
+        });
+
+        console.log(`[tutorial] started: ${tutorialName}`);
+
+        for (let i = 0; i < maxSteps; i++) {
+            const stepInfo = await page.evaluate(() => {
```
james-strauss-uwa, line 78: What does this do? It looks like it builds an object containing info about this tutorial step (currentStep). But it also adds information from the tutSystem object (activeTut, activeTutCurrentStepIndex, activeTutNumSteps), does that info have to be duplicated inside the "info" object?

For example, the "hasActiveTutorial" attribute is used immediately after the object is built, just to abort the  function if necessary. Do we need the "hasActiveTutorial" attribute in the "info" object?

Ah, actually, maybe I understand now. This info gathering step has to be separate since it runs in the browser context using page.evaluate(). Whereas later we use the information to create playwright test steps in the playwright context. Add some comments to make this clearer.

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+                    index: tutSystem?.activeTutCurrentStepIndex ?? -1,
+                    total: tutSystem?.activeTutNumSteps ?? -1,
+                };
+            });
+
+            if (!stepInfo.hasActiveTutorial) {
+                return;
+            }
+
+            const stepDescriptor = `${tutorialName} | ${stepInfo.index + 1}/${stepInfo.total} | ${stepInfo.title} | type=${stepInfo.stepType}`;
```
james-strauss-uwa: This "stepDescriptor" object looks good, but we only use it in one place (line 173). Can we avoid making the object and just build the string if a step fails?

Or if you want to use the stepDescriptor more often, create it once and use it in every log message.

If you don't end up removing the "stepInfo" object, you could add this string as a "descriptor" attribute on that object.

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
         await page.waitForTimeout(500);
     }
 
+    static async waitForTutorialStep(page: Page, title: string): Promise<void> {
+        await page.locator('#tutorialInfoPopUp').waitFor({ state: 'attached', timeout: 10000 });
+        await expect(page.locator('#tutorialInfoPopUp .tutorialInfoTitle h4')).toContainText(title, { timeout: 5000 });
```
james-strauss-uwa: It looks like we use a variety of different lengths for timeouts (1500, 3000, 5000, 7000, 10000).

Maybe standardise on one or two and define them as:

- TestHelpers.SHORT_TIMEOUT
- TestHelpers.LONG_TIMEOUT
- (more if required...)

src/Tutorial.ts
```diff
@@ -879,6 +893,11 @@ export class TutorialStep {
         return this
     }
 
+    setTestStepFunction = (newTestStepFunction:string): this => {
```
james-strauss-uwa: I think I'd prefer the 'newTestStepFunction' to be of type function instead of string.

I guess this would simplify the need to support 3 different "formats" of function strings in the "parseTutorialCustomStepHook" function.

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+                args: args.map((arg) => String(arg)),
+            };
+        }
+
+        // Legacy format: command:arg1:arg2
+        const [command, ...args] = raw.split(':');
+        return { command, args };
+    }
+
+    private static async runTutorialCustomStep(page: Page, testStepFunction: string): Promise<void> {
```
james-strauss-uwa: I hope this function would almost disappear if we moved to use a function instead of a string for the "testStepFunction" parameter.

The different command "sections" of this function would be split out into TestHelpers functions. "centerGraph" is already like that, it just calls TestHelpers.centerGraphOnCanvas().

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+            await canvas.waitFor({ state: 'visible', timeout: 5000 });
+
+            // Right-click around viewport center with pseudo-random offsets and no immediate repeats.
+            const viewport = page.viewportSize();
+            const size = viewport ?? { width: 1280, height: 720 };
+
+            const seed = TestHelpers.contextMenuAnchorIndex;
+            TestHelpers.contextMenuAnchorIndex += 1;
+
+            const angleDeg = (seed * 137.508) % 360;
```
james-strauss-uwa: I might need some extra comments here. I'm not sure why this "openCanvasContextMenu" command doing this angle/radius/circle calculation.

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+        // Shortcut from KeyboardShortcut.ts: center_graph => key "c"
+        await page.press('body', 'c');
+    }
+
+    private static getTutorialInputText(title: string, expectedInput: string): string {
+        if (expectedInput && expectedInput.trim().length > 0) {
+            return expectedInput;
+        }
+
+        // Some tutorial steps ask for a description but don't provide expectedInput.
```
james-strauss-uwa: Could we just return "Automated tutorial text" for all empty expectedInput?

Do we need to treat the "description" case in a special way if this is all automated anyway?

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+        if (!clickPosition) {
+            return false;
+        }
+
+        // Use Playwright's mouse API to dispatch real pointer events.
+        await page.mouse.click(clickPosition.x, clickPosition.y);
+        return true;
+    }
+
+    private static async clickElementByViewportCenter(page: Page, selector: string): Promise<boolean> {
```
james-strauss-uwa: "ViewportCenter" in the function name confused me a little. I think of the viewport as the whole visible page size.

Could we rename this to "clickElementBoundingRectCenter" or something similar.

e2e/TestHelpers.ts
```diff
@@ -21,6 +29,706 @@ export class TestHelpers {
...
+                bubbles: true,
+                cancelable: true,
+            });
+
+            nativeEl.dispatchEvent(keydownEvent);
+            return true;
+        }, value);
+    }
+
+    private static async selectNodeByInterface(page: Page, nodeName: string): Promise<void> {
```
james-strauss-uwa: Would it be better to rename this function selectNodeByName?

It sucks that we need so many (5+) attempts here to reliably select a Node, but I have experienced playwright flaky-ness trying to do similar things, so I understand the need for a complicated function.

Maybe create a JIRA ticket to come back to this later, maybe humans are better at writing this code than AI.

src/main.ts
```diff
@@ -175,7 +177,8 @@ $(function(){
     }
   
     // if not first time visit, but version is newer than last seen version, show the versions modal
-    if (!firstTimeVisit && showWhatsNew && !skipTutorial){
+    // skip if a tutorial was explicitly requested via the url, so it doesn't appear over the tutorial
+    if (!firstTimeVisit && showWhatsNew && !skipTutorial && !tutorialRequested){
```
james-strauss-uwa, line 181: Nice fix!

src/Tutorial.ts
```diff
@@ -765,11 +773,12 @@ export class TutorialStep {
     private preFunc: ((eagle: Eagle) => void) | null;
     private backPreFunc: ((eagle: Eagle) => void) | null;
     private conditionFunc : ((eagle: Eagle) => boolean) | null;
+    private testStepFunction: string | null;
 
     private backSkip : boolean;
     private expectedInput : string;
 
-    constructor(title: string, text: string, type: TutorialStep.Type, waitType: TutorialStep.Wait, delayAmount: number | null, targetFunc: () => JQuery<HTMLElement>, preFunc: ((eagle: Eagle) => void) | null, backPreFunc: ((eagle: Eagle) => void) | null, backSkip:boolean, expectedInput:string, conditionFunc:((eagle: Eagle) => boolean) | null, alternateHighlightTargetFunc: (() => JQuery<HTMLElement>) | null) {
+    constructor(title: string, text: string, type: TutorialStep.Type, waitType: TutorialStep.Wait, delayAmount: number | null, targetFunc: () => JQuery<HTMLElement>, preFunc: ((eagle: Eagle) => void) | null, backPreFunc: ((eagle: Eagle) => void) | null, backSkip:boolean, expectedInput:string, conditionFunc:((eagle: Eagle) => boolean) | null, alternateHighlightTargetFunc: (() => JQuery<HTMLElement>) | null, testStepFunction: string | null) {
```
james-strauss-uwa: I've mentioned this elsewhere, but I guess I'd change the testStepFunction type:

`testStepFunction: ((eagle: Eagle) => void)`

james-strauss-uwa (changes requested):

e2e/TestHelpers.ts
M-Wicenec: this was a way the ai decided to prevent the test from looping endlessly. ive changed it to instead have a 5 attempts per step system.

e2e/TestHelpers.ts
M-Wicenec: this is actually cause there was a bug where while playwright was awaiting, the current step could change, mixing the info from the current and next step. taking the snapshot makes sure that doesnt happen.

ive added a comment

e2e/TestHelpers.ts
M-Wicenec: ive moved it to only be generated when needed by the error catch

e2e/TestHelpers.ts
james-strauss-uwa: Disregard this one

src/Tutorial.ts
james-strauss-uwa: Disregard this one

e2e/TestHelpers.ts
james-strauss-uwa: I couldn't find a way to re-use the TutorialStep.Type enum here. In the end the best I could do was make them both use "string enums" instead of integer enums. This will make the matching slightly more strict.

src/Tutorial.ts
james-strauss-uwa: Disregard the first part of this one.

And you've simplified the "parseTutorialCustomStepHook" code, so I'm happy the second part is done.

e2e/TestHelpers.ts
M-Wicenec: i have rewritten this function to reliably select nodes by using a hybrid approach. playwright will query eagle to find the node's position on graph, and the node's id, which is used as id attribute on nodes elements in the canvas. we can then ask playwright to click that element, or as a backup click at the position of the node.

M-Wicenec:
This is ready for another look

e2e/TestHelpers.ts
james-strauss-uwa: Added EAGLE-1674 to come back to this later

e2e/TestHelpers.ts
james-strauss-uwa: Created EAGLE-1674 to revisit this later

e2e/TestHelpers.ts
james-strauss-uwa: Switched this to "Automated tutorial text."

james-strauss-uwa (approved):

PR #1050, opened 2026-07-02
EAGLE-1670: Warn user of whitespace in branch name

james-strauss-uwa:
Added a validator function to check that user input for branch names satisfy the requirements for git branch names.

Additional validator functions for URLs and non-empty strings. Then updated various places in the code to use the new validator functions where appropriate:

- Enter Graph Name
- Enter Palette Name
- Load File from URL
- Save As
- New Graph Configuration Name
- Create Branch
- Enter Translator URL
- Enter Commit Message

Decided to add branch name validator to both client and server, so the code is once in javascript and once in python.

Added unit tests to check branch name, URL, and non-empty string validation.

## Summary by Sourcery

Introduce validated user input for branch names, URLs, and other text fields, and block invalid branch creations both client- and server-side.

New Features:
- Add reusable client-side validators for non-empty strings, HTTP(S) URLs, and Git-compatible branch names used across multiple input dialogs.
- Enhance the generic text input modal to support inline validation feedback and prevent submission of invalid string input.

Bug Fixes:
- Prevent creation of branches with whitespace or other invalid characters in both the UI and the server API.
- Block empty or invalid values when entering commit messages, configuration names, filenames, URLs, and translator URLs.

Tests:
- Add end-to-end tests covering branch name validation and URL/Save As input validation UX.
- Add helper utilities for reliably closing the input modal during tests.

### DIFF
--- a/e2e/TestHelpers.ts
+++ b/e2e/TestHelpers.ts
@@ -281,11 +281,21 @@ export class TestHelpers {
     }
 
     static async closeInputModalWithoutCompleting(page: Page): Promise<void> {
-        if (await page.locator('#inputModal').isVisible()) {
-            await page.locator('#inputModal button.btn-close').click();
-            await page.waitForTimeout(100);
+        const inputModal = page.locator('#inputModal');
 
-            if (await page.locator('#inputModal').isVisible()) {
+        if (page.isClosed()) {
+            return;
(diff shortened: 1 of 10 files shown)


M-Wicenec (approved):
nice

PR #1056, opened 2026-07-23
EAGLE-1151: Handle empty/malformed graph version safely

james-strauss-uwa:
Updated Utils.determineEagleVersion to handle undefined and empty strings, and fall back to "generatorVersion" is the eagleVersion is not present.

Updated Utils._parseEagleVersion to handle undefined, null, empty string, "unknown", and versions with no "v" prefix.

Fixes hanging bug that used to occur when loading a graph with no "eagleVersion".

## Summary by Sourcery

Improve Eagle version handling to robustly support missing or malformed version metadata and avoid hangs when loading graphs.

Bug Fixes:
- Prevent application hangs when loading graphs that lack a valid eagleVersion field.

Enhancements:
- Make determineEagleVersion resilient to undefined, empty, or missing eagleVersion values by falling back to generatorVersion or a default value.
- Introduce a centralized eagle version parser that safely handles null, undefined, empty, "unknown", and unprefixed version strings before comparing versions.

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -1669,14 +1669,19 @@ export class Utils {
         return Eagle.FileType.Unknown;
     }
 
-    static determineEagleVersion(data: any): string {
-        if (typeof data.modelData !== 'undefined'){
-            const eagleVersion = data.modelData.eagleVersion;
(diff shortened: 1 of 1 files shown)


M-Wicenec (approved):
nice

